PR #1, opened 2026-02-12
update sdk v0.6.0

dzdidi:
Updated to SDK 0.6.0

There is 0.7.0-rc also coming but nothing is on horizon that I envision as a breaking change for the library


ovitrif (approved):
utAck :shipit:

Issue #2, opened 2026-02-12
Re-Export SDK


dzdidi:
For the case of getting profile and social graph from arbitrary app as well enabling for additional unforeseen functional requirements Pubky-SDK should be re-exported

PR #4, opened 2026-02-13
Feat: profile

dzdidi:
Add profile getter and re-exports pubky-sdk for more complicated usacases

https://github.com/pubky/paykit-rs/issues/5
https://github.com/pubky/paykit-rs/issues/2


ovitrif (approved):
utAck

LGTM, added 1 nit but later on wasn't so fond of it, seeing the error has message anyways.

PR #23, opened 2026-02-27
Implement UniFFI mobile bindings for paykit-lib

ben-kaufman:
## Summary

- Adds `paykit-ffi` crate with UniFFI bindings exposing paykit-lib to
  iOS (Swift) and Android (Kotlin), covering profile retrieval, contacts,
  public payment endpoints, authentication via Pubky Ring session import,
  and runtime network switching.
- Includes build scripts, iOS SPM packaging, Android Gradle/Maven
  publishing, and GitHub Actions CI matching bitkit-core's patterns.

## Exported API

| Category | Functions |
|---|---|
| Init | `paykit_initialize`, `paykit_switch_network` |
| Session | `paykit_is_authenticated`, `paykit_get_current_public_key`, `paykit_export_session` |
| Auth | `paykit_import_session`, `paykit_sign_up`, `paykit_sign_in`, `paykit_sign_out`, `paykit_force_sign_out` |
| Reads | `paykit_get_profile`, `paykit_get_contacts`, `paykit_get_payment_list`, `paykit_get_payment_endpoint` |
| Writes | `paykit_set_payment_endpoint`, `paykit_remove_payment_endpoint` |


paykit-ffi/bindings/android/lib/src/main/jniLibs/arm64-v8a/libpaykit.so
dzdidi: would it make sense to include these `*.so` files into gitignore and let CI handle them?

ben-kaufman:
Thanks, I think I've answered/ fixed for all points in the specific comments and in the commit I pushed.

Just about how to organize this, I think for now at least the separation seems cleaner the way it is, but if you think it'll be better for the future we could move it easily. Let me know what you think.

PR #29, opened 2026-03-06
feat: remove sdk

dzdidi:
## Summary

- Removed `get_profile` and `get_known_contacts` helpers — callers should use the Pubky SDK directly.
- Removed `pub use pubky` re-export from `paykit-lib`.
- Removed `PaykitError::Profile` variant, `PUBKY_FOLLOWS_PATH` constant, and `pubky-app-specs` dependency.
- Removed FFI bindings: `paykit_get_profile`, `paykit_get_contacts`, `FfiProfile`, `FfiProfileLink`, `PaykitFfiError::ProfileError`.

## Release

[v0.1.0-rc2](https://github.com/pubky/paykit-rs/releases/tag/v0.1.0-rc2)


ovitrif (changes requested):
@dzdidi
Reviewed manually, lgtm with one remark: we must version for the bindings, as bitkit-android and -ios already use an older version.

We can use `0.1.0-rc$N` where `$N` is autoincrement index.

To make better sense of progress, I am planning to also split changelog by these `-rc$N` releases.

Let me know if I'm missing something and this is not very well aligned with what we've been discussing on 2026-03-10. Thing is, once you start using a library as a dependency, and we did, things get messy if we don't version.

---

PS. the request changes is for me to version and update changelog, just added it as a means of stronger signalling for the above comments.

PR #18, opened 2026-02-25
feat: error source

dzdidi:
## Pull request overview

https://github.com/pubky/paykit-rs/issues/10

This pull request refactors the `PaykitError` enum to use structured error variants with proper error source chain preservation. The changes introduce `thiserror` and `anyhow` dependencies to modernize error handling, moving from simple string-based error messages to structured errors that preserve the underlying error causes.

**Changes:**
- Refactored `PaykitError` enum to use `thiserror::Error` derive macro and structured variants with `context` and `source` fields
- Updated all error construction sites to use the new structured error format with preserved error sources via `anyhow::Error`
- Removed the unused `Unimplemented` error variant and manual `Display`/`Error` trait implementations

### Reviewed changes

Copilot reviewed 4 out of 5 changed files in this pull request and generated no comments.

<details>
<summary>Show a summary per file</summary>

| File | Description |
| ---- | ----------- |
| paykit-lib/src/lib.rs | Refactored `PaykitError` to use `thiserror` derive macro, converted `Transport` and `InvalidData` to struct variants with source preservation, removed manual trait implementations |
| paykit-lib/src/transport/pubky/unauthenticated_transport.rs | Updated all error construction sites to use new structured error format with context and source fields |
| paykit-lib/src/transport/pubky/authenticated_transport.rs | Updated error construction sites to use new structured error format |
| paykit-lib/Cargo.toml | Added `anyhow` and `thiserror` dependencies |
| Cargo.lock | Updated lockfile with new dependencies and version updates for `anyhow` and `thiserror` |

Issue #13, opened 2026-02-13
Race Condition in fetch_supported_payments (unauthenticated_transport.rs:80-110)


dzdidi:
```
     let entries = self.list_entries(addr, "list supported payments").await?;
     // ...
     for resource in entries {
         // Fetch each endpoint individually
         if let Some(payload) = self.fetch_text(resource.to_string(), &label).await? {
```
 Problem: Between listing and fetching, endpoints can be added/removed/modified. The returned SupportedPayments may be inconsistent.

Recommendation: Document this limitation clearly. Consider adding a fetch_supported_payments_snapshot that uses atomic operations if the storage layer supports it.

Issue #12, opened 2026-02-13
fetch_known_contacts Fails on Single Invalid Entry (unauthenticated_transport.rs:142-146)


dzdidi:
```
     Err(err) => {
         return Err(PaykitError::Transport(format!(
             "invalid contact entry '{pk_str}': {err}"
         )))
     }
```
Problem: If one contact has an invalid public key format, the entire operation fails. This is fragile—a single corrupted entry breaks contact discovery.

Recommendation: Log/skip invalid entries and continue:
```
     Err(_) => {
         // Log warning, skip invalid entry
         continue;
     }
```

Issue #10, opened 2026-02-13
Error Type Lacks source() Chain (lib.rs:76-111)


dzdidi:
```
     #[derive(Debug)]
     pub enum PaykitError {
         Transport(String),  // Loses the original error!
         // ...
     }
```
Problem: Original errors are stringified, losing:
     - Error chain for debugging
     - Ability to downcast to specific error types
     - Structured error handling by callers

Recommendation:
```
     #[derive(Debug)]
     pub enum PaykitError {
         Transport {
             context: &'static str,
             source: Box<dyn std::error::Error + Send + Sync>,
         },
         // ...
     }
```
Or use `thiserror` crate for ergonomic error handling.

PR #30, opened 2026-03-11
feat: private payments

dzdidi:
Adding private payments

- Integrates noise library for private payments
    - handshake management
    - transport handling
- updated dependency
- updated docs

No bindings has been included in this PR as I am not on mac. Either can be included by somebody else into this PR or in a separate one


ovitrif:
@dzdidi why does this still has 'draft' in its title?!
because draft = not PR draft, but like, first pre-MVP iteration?!

dzdidi:
@ovitrif good catch, thank you, it is quite final from my assessment unless review proves me wrong

paykit-lib/README.md
```diff
@@ -1,62 +1,105 @@
...
+    println!("{}: {}", method.as_str(), data.as_str());
+}
+```
 
 ## Core Types
 
 ### `MethodId`
 
-Identifier for a payment method specification (e.g. `"lightning"`, `"onchain"`, `"bolt11"`).
+Identifier for a payment method specification (e.g. `"onchain"`, `"bolt11"` basically anything parties agree on).
```
ben-kaufman: nit and not relevant to the PR but "onchain" is very broad as an endpoint name (which chain, which address type etc.) maybe better write "bitcoin-p2tr", "bitcoin-p2wpkh" etc.

paykit-lib/README.md
```diff
@@ -1,62 +1,105 @@
...
+use paykit_lib::EndpointData;
+
+let data = EndpointData::new("ln...");
 let payload: &str = data.as_str();
 let owned: String = data.into_inner();
 ```
 
+### `SupportedPayments`
+
+Collection of payment entries keyed by method identifiers. Returned by `get_payment_list` and `get_private_payments`.
```
ben-kaufman, line 83: nit: Could have the example cover get_private_payments as well

paykit-lib/README.md
```diff
@@ -68,18 +111,264 @@ Domain error enum with the following variants:
...
+
+    let method = MethodId::new("bolt11")?;
+    let data = EndpointData::new("ln...");
+    set_payment_endpoint(client, method, data).await?;
+
+    let method = MethodId::new("p2wpkh")?;
+    let data = EndpointData::new("bc1...");
+    set_payment_endpoint(client, method, data).await?;
+    // or 
+    let method = MethodId::new("onchain")?;
```
ben-kaufman: nit and unrelated directly to the PR, but would it make sense to rename this or have some list of set conventions already? I think that's an important part of the protocol as if names are not fully defined and agreed on the protocol will simply not fulfill its purpose

paykit-lib/README.md
```diff
@@ -68,18 +111,264 @@ Domain error enum with the following variants:
...
+
+#### Handshake Initiation
+- `initiate_encrypted_link(session, sender_secret_key, receiver_pubkey, outbox_client) -> Result<EncryptedLinkHandshake>`  
+  Initializes a Noise XX handshake as the **initiator**. Returns a handshake handle to be driven forward with `advance_handshake`.
+- `accept_encrypted_link(session, receiver_secret_key, sender_pubkey, outbox_client) -> Result<EncryptedLinkHandshake>`  
+  Initializes a Noise XX handshake as the **responder**. Returns a handshake handle to be driven forward with `advance_handshake`.
+
+**NOTE**: Due to nature of noise it is important that one of the peers is "initiator" and another is "responder". Sometimes it is impossible to determine who is who based on user flow. One option is to compare counterparty key to own key let initiator be the one with lexicographically bigger public key.
```
ben-kaufman:
```suggestion
**NOTE**: Due to nature of noise it is important that one of the peers is "initiator" and another is "responder". Sometimes it is impossible to determine who is who based on user flow. One option is to compare counterparty key to own key and let the initiator be the one with lexicographically bigger public key.
```

ben-kaufman (commented):
Note that the repo level README.md needs significat update on how private payments are expected to work (among potentially other points).

paykit-lib/src/lib.rs
```diff
@@ -245,6 +241,321 @@ pub struct SupportedPayments {
...
+    /// The snapshot contains everything needed to restore the session later
+    /// via [`restore_encrypted_link`] or [`restore_encrypted_link_from_config`]
+    /// without re-doing the Noise handshake.
+    ///
+    /// # When to snapshot
+    ///
+    /// Take a snapshot after the link is established and periodically after
+    /// exchanging messages (the snapshot includes nonce counters that must stay
+    /// in sync). Persist the serialized bytes to durable storage so the session
+    /// can be resumed after an app restart.
```
ben-kaufman: I think we should explicitly mention here that the serialized snapshot is sensitive material, not just resume metadata. As far as I understand the underlying pubky-noise session state includes secret key material too, so "persist to durable storage" should probably mention it must be treated like a secret and stored encrypted/ in secured storage at rest.

paykit-lib/src/lib.rs
```diff
@@ -322,6 +714,47 @@ where
...
+/// - Returns `Err(PaykitError::Transport)` for decryption or I/O failures.
+#[instrument(skip(link), fields(recipient = %link.recipient))]
+pub async fn get_private_payments(link: &mut EncryptedLink) -> Result<SupportedPayments> {
+    debug!("receiving private payments map");
+
+    let messages = link.encryptor.receive_message().await;
+
+    if messages.is_empty() {
+        debug!("no private payments messages available, returning empty map");
+        return Ok(SupportedPayments::default());
```
ben-kaufman: One thing I am not sure about here: does empty from `receive_message` really mean only "no new message yet"? It looks like upstream can also return empty on some read/decode failure cases, so this may hide transport/corruption problems as just an empty map.

paykit-lib/src/lib.rs
```diff
@@ -270,6 +581,87 @@ where
...
+        .map_err(|err| map_error("set_private_payments", err))?;
+
+    let plaintext = json.into_bytes();
+
+    if plaintext.len() > pubky_noise::snow_crypto::PUBKY_NOISE_MSG_LEN {
+        return Err(PaykitError::Validation(format!(
+            "private payments payload ({} bytes) exceeds max message size ({} bytes)",
+            plaintext.len(),
+            pubky_noise::snow_crypto::PUBKY_NOISE_MSG_LEN,
+        )));
```
ben-kaufman, line 762: I think this size check is a bit too loose. The Noise transport still needs room for the AEAD tag, so the real max plaintext should be below 1000 bytes (should reduce 16 for the AEAD tag). Right now payloads close to 1000 can pass validation here and then fail later as transport errors after retries.

paykit-lib/src/lib.rs
```diff
@@ -322,6 +714,47 @@ where
...
+
+    let messages = link.encryptor.receive_message().await;
+
+    if messages.is_empty() {
+        debug!("no private payments messages available, returning empty map");
+        return Ok(SupportedPayments::default());
+    }
+
+    // Take the last message (latest state of the payments map).
+    let raw = &messages[messages.len() - 1];
```
ben-kaufman: I'm not sure this really returns the "latest state" if multiple updates are already queued. Since receive_message is counter/slot based, the first call still returns the first unread update, not necessarily the newest one. Should this helper maybe drain unread slots and return the latest map, or otherwise be documented more as queue semantics?

paykit-lib/src/lib.rs
```diff
@@ -361,6 +794,449 @@ where
...
+
+    let config = pubky_noise::PubkyNoiseConfig::new_with_paths(
+        sender_secret_key,
+        0,
+        "XX",
+        session,
+        write_path,
+        read_path,
+        outbox_client,
+    )
```
ben-kaufman, line 994: Should we validate here that the provided session actually matches the provided secret key? Right now if a caller mixes them by mistake, paths/Noise state are derived from one identity while writes happen with another session, and that will then fail later in a pretty confusing way instead of giving a direct validation error.

paykit-lib/README.md
dzdidi: Could indeed be a better approach

paykit-lib/README.md
dzdidi: My bet was that these conventions will emerge naturally from implementations. More info here https://github.com/pubky/paykit-rs?tab=readme-ov-file#paykit-method-implementation-proposals

ben-kaufman: I see, I'd think actually the opposite, having more like a centralize list for the "official" protocol conventions would be better, ie. not binding names on the protocol levels but just the "standard" which should be the basis everyone can easily refer to. But since this is completely unrelated to the PR we can discuss it later elsewhere :)

paykit-lib/src/lib.rs
dzdidi: > the first call still returns the first unread update, not necessarily the newest one.

this is correct. 


Draining is the right approach here because counters are global

paykit-lib/src/lib.rs
dzdidi: > Should we validate here that the provided session actually matches the provided secret key?

this is not strictly a requirement, by itself. 

The session was intended to be created by pubky-ring and corresponding private key should not leave it. The `sender_secret_key` is not necessarily related to session but it is correct that it needs to be consistent through restoration/reinstantiation


I think they should be two different keys. As session key is the one responsible managing homeserver auth + pkarr identity. While `sender_secret_key` is only used in scope of the paykit.


So it is  `root_key` for session and `application_key` for paykit

paykit-lib/src/lib.rs
ben-kaufman: This can also cause get_private_payments to stop draining too early

paykit-lib/src/lib.rs
```diff
@@ -361,63 +935,657 @@ where
...
+    snapshot: EncryptedLinkSnapshot,
+) -> Result<EncryptedLink> {
+    if snapshot.recipient() != remote_pubkey {
+        return Err(PaykitError::Validation(format!(
+            "remote_pubkey does not match snapshot recipient (remote={}, snapshot={})",
+            remote_pubkey,
+            snapshot.recipient(),
+        )));
+    }
+
```
ben-kaufman, line 1522: Could add here state validation similar to `restore_encrypted_link_handshake_inner`
```
if !matches!(
         snapshot.state.phase,
         pubky_noise::snow_crypto::NoisePhase::Transport
     ) {
         return Err(PaykitError::Validation(format!(
             "encrypted link restore requires transport-phase snapshot, got {:?}",
             snapshot.state.phase,
         )));
     }
```
Otherwise it seems possible to pass a handshake snapshot here, get back an EncryptedLink, and only fail later in a confusing way when send/receive are used.

paykit-lib/src/lib.rs
```diff
@@ -270,6 +709,87 @@ where
...
+/// - `link` — an established [`EncryptedLink`] for encryption and I/O.
+/// - `entries` — the complete map of payment methods to store.
+///
+/// # Errors
+/// - Returns [`PaykitError::Validation`] if the serialized payload exceeds
+///   the maximum message size.
+/// - Returns [`PaykitError::InvalidData`] if the map cannot be serialized.
+/// - Returns [`PaykitError::Transport`] if `send_message` fails after all
+///   retry attempts are exhausted.
+#[instrument(skip(link, entries), fields(recipient = %link.recipient, count = entries.len()))]
```
ben-kaufman: Having recipient here exposes the counterparty pubkey in tracing/logs, which can be a privacy concern. I’d avoid logging it unless really needed.

paykit-lib/src/lib.rs
dzdidi: 1000 bytes is an arbitrarily chosen constant for restriction of the input message size inside of pubky-noise, the 2 bytes for payload length and 16 bytes for AEAD tag are added on top of these 1000 bytes, making it 1018bytes which is still significantly less than the Noise protocol message restriction which is 65535 bytes

paykit-lib/src/lib.rs
dzdidi: What do you mean?

paykit-lib/src/lib.rs
ben-kaufman: Still, the issue I meant is that some payloads can pass validation here and then fail later in the Noise layer, so there still seems to be a mismatch between the limit checked in Paykit and the limit actually accepted downstream. Also I think the 65535 Noise protocol max is not really the relevant limit here, since in the current pubky-noise implementation there is still a fixed 1000-byte message buffer being used.
The main point is that there seems to be a slight inconsistency between the limit here and in pubky-noise, which should probably be adjusted.

paykit-lib/src/lib.rs
ben-kaufman: Fix looks good, but may also make sense then to remove it from the handshake process methods too (ie initiate_encrypted_link, accept_encrypted_link etc.)

paykit-lib/README.md
dzdidi: related to https://github.com/pubky/paykit-rs/pull/30#discussion_r3039432666 

For docs writing `"bitcoin-p2tr"` etc makes perfect sense to me.  For the general usage I postpone this decision

paykit-lib/src/lib.rs
ben-kaufman: What I mean is that `get_private_payments` treats `receive_message().is_empty()` as "there are no more unread messages". But as far as I can tell, upstream `receive_message() `can also return empty on some read/decode failure paths, not only when there is truly no new message.

So this can affect the draining logic too:

- If the first read fails and returns empty, this returns an empty map as if nothing was sent
- If one update is read successfully, and then a later unread slot fails and returns empty, the loop stops early and returns the previous update as if it were the latest one

So the issue is that an empty result here may hide an actual transport/corruption problem and can also make draining stop too early.

paykit-lib/src/lib.rs
dzdidi: > Still, the issue I meant is that some payloads can pass validation here and then fail later in the Noise layer,

I do not see how this is possible

paykit-lib/src/lib.rs
dzdidi: good catch, thank you

dzdidi: Removed logging of counterparty's public key from everywhere

paykit-lib/src/lib.rs
dzdidi: If read of message `N-1` fails, read of  message `N` will never succeed because of noise state and keys derived from it. 


> But as far as I can tell, upstream receive_message() can also return empty on some read/decode failure paths, not only when there is truly no new message.

This will be addressed in pubky-noisehttps://github.com/pubky/pubky-noise/issues/8

paykit-lib/src/lib.rs
ben-kaufman: I wrote a tiny local repro to verify this more directly. The test I used was:

```
#[cfg(test)]
mod tests {
    use pubky_noise::snow_crypto::PUBKY_NOISE_MSG_LEN;
    use snow::{params::NoiseParams, Builder};

    #[test]
    fn payload_boundary_repro_for_paykit_vs_downstream_limits() {
        let passes_paykit_but_should_fail_downstream = PUBKY_NOISE_MSG_LEN - 15; // 985
        let paykit_rejected = PUBKY_NOISE_MSG_LEN + 1; // 1001

        assert!(
            passes_paykit_but_should_fail_downstream <= PUBKY_NOISE_MSG_LEN,
            "985-byte payload should pass the Paykit-side <= 1000 length check"
        );
        assert!(
            paykit_rejected > PUBKY_NOISE_MSG_LEN,
            "1001-byte payload should fail the Paykit-side check"
        );

        let params: NoiseParams = "Noise_NN_25519_ChaChaPoly_BLAKE2s"
            .parse()
            .expect("valid noise params");
        let mut initiator = Builder::new(params.clone())
            .build_initiator()
            .expect("initiator");
        let mut responder = Builder::new(params).build_responder().expect("responder");

        let mut handshake_buf = [0u8; 1024];
        let len = initiator
            .write_message(&[], &mut handshake_buf)
            .expect("initiator handshake write");
        responder
            .read_message(&handshake_buf[..len], &mut [])
            .expect("responder handshake read");
        let len = responder
            .write_message(&[], &mut handshake_buf)
            .expect("responder handshake write");
        initiator
            .read_message(&handshake_buf[..len], &mut [])
            .expect("initiator handshake read");

        let mut transport = initiator
            .into_transport_mode()
            .expect("transition to transport mode");

        let ok_payload = vec![0u8; PUBKY_NOISE_MSG_LEN - 16];
        let payload_985 = vec![0u8; passes_paykit_but_should_fail_downstream];
        let payload_1001 = vec![0u8; paykit_rejected];
        let mut out = [0u8; PUBKY_NOISE_MSG_LEN];

        assert!(
            transport.write_message(&ok_payload, &mut out).is_ok(),
            "984-byte payload should fit in the 1000-byte buffer"
        );
        assert!(
            transport.write_message(&payload_985, &mut out).is_err(),
            "985-byte payload should fail because the 16-byte AEAD tag must fit in the same buffer"
        );
        assert!(
            transport.write_message(&payload_1001, &mut out).is_err(),
            "1001-byte payload should also fail downstream"
        );
    }
}
```
What this is proving is:
- 985 <= 1000, so a 985-byte payload would pass the current Paykit-side validation
- once you reach transport mode and try to write into the same fixed 1000-byte buffer size used downstream, 984 bytes succeeds but 985 bytes fails
- so there really is a range of payload sizes that pass validation here and still fail later downstream

This is not a full Paykit end-to-end test, but I think it does prove the exact boundary condition the current downstream path relies on.

dzdidi: I see, the buffer being of insufficient size. Thank you

dzdidi: Will be addressed in next RC of pubky-noise https://github.com/pubky/pubky-noise/pull/10

paykit-lib/src/lib.rs
ben-kaufman: It still seems to swallow failures here in [pubky-noise receive_message (line 484).](https://github.com/pubky/pubky-noise/blob/c8f992a5f59b27f458a1c7b2502c12d4f5d830a6/pubky-noise/src/lib.rs#L484)

So this still hides errors in two cases:

if public_storage().get(...) itself returns Err(...)
if get(...) returns Ok(response) but with a non-success status
In both cases it just falls through and returns empty as if there was no message.

I think the fix should be to handle that explicitly and only treat 404 / not found as “no message yet”, while other read / HTTP failures should return Err(...) instead of Ok(empty).

ben-kaufman (approved):

paykit-lib/src/lib.rs
dzdidi: My bad, missed it on the last iteration. Thank you

ovitrif (changes requested):
# PR #30 Review — Private Payments with Noise Encryption

## Critical Issues (2 found)

**1. Original error type stringified, breaking downcast contract**
- `paykit-lib/src/lib.rs:774-798` — In `set_private_payments`, the `PubkyNoiseError` from `send_message` is immediately converted to a string via `format!("{err:?}")`. The final `PaykitError::Transport` source is an `anyhow::anyhow!()` wrapping that string, destroying the typed error. The crate's own `PaykitError` doc says callers can `downcast_ref` the source — this violates that contract.

**2. All-zero payload produces misleading JSON parse error**
- `paykit-lib/src/lib.rs:900` — If `receive_message` returns a buffer of all zeros, the zero-trimming logic produces an empty slice, which then fails at `serde_json::from_str("")` with "EOF while parsing a value". Operators will chase JSON serialization bugs when the real problem is a corrupted/empty encrypted payload. Add an explicit check for `end == 0` returning `PaykitError::InvalidData` with a clear message about zero-byte payload.

## Important Issues (8 found)

**3. `serialize_private_payments_json` uses wrong error variant**
- `paykit-lib/src/lib.rs:676-679` — Uses `PaykitError::InvalidData` for serializing caller-supplied data. Per CLAUDE.md: `InvalidData` is for network-fetched corrupt data; `Validation` is for caller input failures.

**4. Mid-drain receive error silently loses already-consumed messages**
- `paykit-lib/src/lib.rs:877-892` — `get_private_payments` drain loop: if `receive_message()` errors after successfully draining earlier batches, the `?` propagates immediately and all previously drained messages are irrecoverably lost. At minimum, log a warning with the drained count before returning the error.

**5. `source: None` in snapshot deserialize discards error chain**
- `paykit-lib/src/lib.rs:398-405, 470-479` — Both `EncryptedLinkSnapshot::deserialize` and `EncryptedLinkHandshakeSnapshot::deserialize` set `source: None`, breaking the `std::error::Error::source()` chain. If the upstream error type supports it, use `Some(err.into())`.

**6. Last retry attempt not logged**
- `paykit-lib/src/lib.rs:774-783` — The `if attempt < max_attempts` condition skips logging on the final failed attempt. The most important failure gets no structured log entry.

**7. Snapshot types don't enforce phase invariant at deserialization**
- `EncryptedLinkSnapshot::deserialize` accepts handshake-phase bytes without error (and vice versa). Phase validation only happens later at the restore callsite. Adding phase checks in `deserialize` would make the types genuinely distinct and catch misuse earlier.

**8. Missing test: encrypted link restore rejecting handshake-phase snapshot**
- The converse test exists (handshake restore rejects transport-phase), but there's no test verifying `restore_encrypted_link_from_config` rejects a handshake-phase snapshot. Low-effort, high-value addition.

**9. Missing tests for `parse_private_payments_json` with adversarial input**
- No tests for invalid JSON or invalid MethodId keys (e.g., `{"../attack": "value"}`). CLAUDE.md requires "full coverage on serialization/deserialization paths that map to network data."

**10. CLAUDE.md/AGENTS.md missing "private" reserved word for MethodId**
- `CLAUDE.md:54` — The MethodId validation docs don't mention that `"private"` is reserved.

## Suggestions (9 found)

**11. `lib.rs` has grown to 2,726 lines** — Consider extracting private payment types/functions into a `src/private.rs` or `src/noise.rs` module.

**12. Typo in README** — `paykit-lib/README.md:174`: `"bitcoint-bolt11"` should be `"bitcoin-bolt11"`.

**13. Inconsistent field naming** — `EncryptedLink.recipient` vs `EncryptedLinkHandshake.remote_pubkey` for the same concept.

**14. Missing `recipient()` accessor** — Neither `EncryptedLink` nor `EncryptedLinkHandshake` expose the counterparty pubkey directly; callers must go through `snapshot().recipient()`.

**15. `close_encrypted_link` is async but performs no async work** — Consider making it sync, or documenting the async signature as forward-compatible.

**16. `max_send_retries` silently reset on restore** — Unlike the handshake type, the docs don't mention this.

**17. Plain `+= 1` on recovery counter** — `lib.rs:1168` uses plain addition instead of `saturating_add`, inconsistent with the project's own safety pattern (see commit "safety: saturated increment to avoid overflow").

**18. `compute_private_payment_paths` doc uses nonexistent `.write`/`.read` accessors** — Function returns a tuple, not a struct. Should reference `.0`/`.1`.

**19. Non-ASCII en-dash in doc comment** — `lib.rs:348` uses an en-dash while line 437 uses `-`. Standardize on ASCII per project style.

## Strengths

- **Build is clean**: `cargo fmt`, `cargo clippy`, and `cargo doc` all pass with no warnings
- **Security-conscious**: Secret keys properly `skip`-ed in all `#[instrument]` annotations; snapshot `Debug` impls redact sensitive state; doc comments warn about treating snapshot bytes as secrets
- **Excellent type design**: The `HandshakeProgress` enum with move semantics is a textbook Rust state machine
- **Thorough test coverage**: 15+ new tests covering happy paths, failure paths, concurrent writer/reader, snapshot round-trips, phase/pubkey mismatch rejection
- **Documentation is high quality**: 189-byte claim verified correct against upstream source; retry/recovery behavior precisely matches implementation; all doc-tests pass

## Recommended Action

1. Fix critical issues #1-2 (error chain preservation, zero-payload handling)
2. Address important issues #3-6 (error variant, mid-drain loss, source chain, logging)
3. Add missing tests #8-9 (low effort, high value)
4. Fix typo #12 and doc inaccuracies #18-19
5. Consider suggestions #7, #10-11, #13-17 as follow-up improvements

ovitrif:
Codex roasted Claude's review so now we're refining the final review + verdict.

ovitrif (approved):
# Codex Review of [Claude's Review](https://github.com/pubky/paykit-rs/pull/30#pullrequestreview-4081912306)

Validated against PR head `99d5c993752f29ae27dbf344aa5bc621b632c08f`.

Build status on current head:

- `cargo fmt --check` passes
- `cargo clippy --all-targets --all-features` passes
- `cargo doc --no-deps` passes
- `cargo test` passes

## Relevant Points

### 1. Zero-payload handling in `get_private_payments`

In `get_private_payments`, trailing zero trimming can reduce a corrupted payload to an empty slice before JSON parsing. In that case the function currently returns a generic JSON EOF error instead of a clearer `PaykitError::InvalidData` describing an empty/corrupt decrypted payload.

File:

- `paykit-lib/src/lib.rs` around `end = raw.iter().rposition(...)`

Why it matters:

- This is a real robustness issue.
- The current error message is misleading during debugging.

Suggested disposition:

- Worth fixing before or after merge.
- Not strong enough on its own to justify `Request changes`.

### 2. Drain-loop behavior on mid-stream receive failure

`get_private_payments` drains unread messages and keeps only the latest map. If one or more messages are consumed successfully and a later `receive_message()` call fails, the function returns an error after already advancing the queue.

File:

- `paykit-lib/src/lib.rs` in the `loop` inside `get_private_payments`

Why it matters:

- This is a real edge case in queue semantics.
- The practical fix is not just "log a warning"; the API behavior itself needs a decision.

Suggested disposition:

- Worth discussing or documenting explicitly.
- Better treated as a follow-up design/robustness item than a merge blocker.

### 3. Missing adversarial parser tests

The private-payments parser does not currently have focused tests for malformed JSON or invalid method IDs inside otherwise valid JSON payloads.

Files:

- `paykit-lib/src/lib.rs` for `parse_private_payments_json`
- existing private-payment tests near the bottom of `paykit-lib/src/lib.rs`

Why it matters:

- The repository guidance asks for strong coverage on serialization/deserialization paths tied to network data.

Suggested disposition:

- Good follow-up.
- Not blocking.

### 4. Small doc fixes worth taking

These are valid but non-blocking:

- `AGENTS.md` and `CLAUDE.md` should mention that `private` is a reserved `MethodId`.
- `paykit-lib/README.md` has `bitcoint-bolt11` instead of `bitcoin-bolt11`.
- `compute_private_payment_paths` docs describe `.write` / `.read` accessors even though the function returns a tuple.
- One doc comment still uses a non-ASCII dash despite the repository's ASCII guidance.
- The encrypted-link restore docs should mention that `max_send_retries` resets to the default on restore.

dzdidi:
> 1. Zero-payload handling in get_private_payments

I might be misunderstanding something in report but if there is `""` after trailing zeros removal, then it will be `parse_private_payments_json("")` , right? But in this case it will be `PaykitError`:
```rust
    #[test]
    fn test_parse_private_payments_json_empty_string() {
        let err = parse_private_payments_json("").unwrap_err();
        assert!(
            matches!(err, PaykitError::InvalidData { ref context, .. } if context.contains("failed to parse")),
            "expected InvalidData parse error for empty string, got: {err}"
        );
    }
```

dzdidi:
> 2.  Drain-loop behavior on mid-stream receive failure

Just retry, error is error. Only latest message matter anyway. Also because of noise, one always needs to read all messages

PR #35, opened 2026-04-21
docs: add payment endpoint identifier spec draft

ben-kaufman:
## Summary
- Adds `specs/payment-endpoint-identifier.md` (Draft v0.1), a recommended but non-obligatory convention for naming payment endpoints as three-segment `asset-rail-endpoint` strings, and the shape of the JSON payload that accompanies them.
- Explicitly ties the spec to the `MethodId` validation in [paykit-lib/src/lib.rs:129](paykit-lib/src/lib.rs) and notes where the spec is stricter than what the library enforces.
- Includes grammar (ABNF), segment rules, valid/invalid examples, payload conventions (`value` required; `min`/`max` and endpoint-specific fields recommended), extensibility notes, and a security considerations section.

## Notes for reviewers
- The document is marked Draft v0.1 and not normative: the intent is to publish an initial officially recommended convention that downstream bindings (Swift/Kotlin/RN) can align on, not to freeze anything.
- Single-canonical-address rails use `address` (`usdt-ethereum-address`, `usdt-tron-address`, `usdc-solana-address`). Rails with genuinely multiple address formats keep specific names (Bitcoin's `p2tr`, `p2wpkh`, etc.).
- No code changes; `MethodId` behaviour is unchanged.

## Test plan
- [ ] Render `specs/payment-endpoint-identifier.md` on GitHub and confirm tables, ABNF block, and links (README, AGENTS.md, paykit-lib/src/lib.rs) resolve.
- [ ] Confirm the examples in §6.1 all pass `MethodId::new` and the examples in §6.2 would fail structural validation or the conformance rules in §4.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/specs/payment-endpoint-identifier.md
+++ b/specs/payment-endpoint-identifier.md
@@ -1,0 +1,268 @@
+# Paykit Payment Endpoint Identifier Specification
+
+| Field       | Value                                      |
+| ----------- | ------------------------------------------ |
+| Status      | Draft                                      |
+| Version     | 0.1                                        |
+| Last updated| 2026-04-21                                 |
+| Repository  | <https://github.com/pubky/paykit-rs>       |
+
+## Abstract
+
+This document specifies a naming convention for *payment endpoint identifiers*
+in Paykit: short, structured strings that name an unambiguous way for a payee
+to receive value. It also defines the shape of the JSON payload that
+accompanies each identifier.
+
+The intent is to give implementers and downstream bindings (Swift, Kotlin,
+React Native) a single, consistent vocabulary to exchange, so that a payer
+and payee who independently support "`bitcoin-lightning-bolt12`" can recognise
+each other without side-channel coordination.
+
+## 1. Status of this document
+
+This is a *recommended* convention, not a mandatory one. Paykit's routing
+layer does not enforce identifier shape beyond the structural checks performed
+by [`MethodId`](../paykit-lib/src/lib.rs) (ASCII alphanumeric, `-`, `_`, `.`;
+max 64 characters; no path traversal; the value `private` is reserved).
+
+Identifiers that follow this specification are interoperable with Paykit
+clients that follow the same conventions. Identifiers that do not follow it
+may still be valid `MethodId` values, but carry no cross-implementation
+guarantees.
+
+A future revision may introduce a formal registry, additional conformance
+checks, or normative validators. Until then, the grammar and conventions
+below are the working agreement.
+
+## 2. Conformance
+
+The key words **MUST**, **MUST NOT**, **SHOULD**, **SHOULD NOT**, and **MAY**
+in this document are to be interpreted as described in [RFC 2119] when, and
+only when, they appear in all capitals.
+
+## 3. Terminology
+
+- **Payment endpoint identifier** (or *identifier*): the three-segment string
+  that names a payment endpoint type, e.g. `bitcoin-onchain-p2tr`. In the
+  Paykit library this is represented by the `MethodId` type.
+- **Payload**: the JSON object stored alongside the identifier, describing
+  the specific payee handle for an endpoint of that type.
+- **Asset**, **rail**, **endpoint**: the three positional segments of an
+  identifier, defined in Section 5.
+- **Segment**: one positional component of an identifier, delimited by `-`.
+
+Paykit's higher-level terms (*Payment Method*, *Payment Endpoint*,
+*Supported Payments List*) are defined in the top-level [README](../README.md);
+this specification does not redefine them.
+
+## 4. Grammar
+
+An identifier is a three-segment string of the form `asset-rail-endpoint`.
+
+```abnf
+identifier   = asset "-" rail "-" endpoint
+asset        = 1*segment-char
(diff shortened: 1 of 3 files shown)


specs/payment-endpoint-identifier.md
```diff
@@ -0,0 +1,268 @@
...
+4. Identifiers are compared byte-for-byte. Implementations MUST NOT perform
+   case folding, Unicode normalisation, or alias resolution when matching.
+
+## 5. Segments
+
+### 5.1 Asset
+
+The *asset* segment names the unit of value being transferred.
+
+- For cryptoassets, use the lowercased ticker (`bitcoin`, `usdt`, `eth`).
```
dzdidi: This seem to be a bit inconsistent. Bitcoin is either `btc` or `xbt`. 

Maybe better to use "asset name" instead of ticker?

Then it should be something like (`bitcoin`, `usdt`, `ether`).

specs/payment-endpoint-identifier.md
```diff
@@ -0,0 +1,268 @@
...
+- For fiat, use the lowercased [ISO 4217] code (`usd`, `eur`, `gbp`).
+- The asset segment MUST NOT encode the chain or token standard; the rail
+  segment disambiguates those.
+
+### 5.2 Rail
+
+The *rail* segment names the settlement system carrying the asset. Examples:
+
+- Base chain: `onchain`, `ethereum`, `solana`, `tron`.
```
dzdidi: Not sure about `onchain` here. For me `bitcoin` seem to be more factually correct. It should also be able to accommodate some smart-contract networks on bitcoin as well. So in the end (with application of comment above) it all will be looking like  `btc-bitcoin-...`

specs/payment-endpoint-identifier.md
```diff
@@ -0,0 +1,268 @@
...
+```
+
+```json
+// eur-sepa-iban
+{
+  "value": "DE89370400440532013000",
+  "bic": "COBADEFFXXX",
+  "beneficiary_name": "Jane Doe"
+}
+```
```
dzdidi, line 210: I am not sure about use of `value` here. I understand that it is coming from attempt of having generic enough term but it makes it hard to discuss this concept specifically because of the overlap with term "value" in the context of JSON being "key-value" store. For example I have a question "Is there restriction on the type of value in payload?" I see restriction on the value of the the key "value" but not on general value itself. Things like SWIFT and ACH payments can require quite complex payloads. If you can add example for both SWIFT and ACH which would also include routing intermediary banks it would be great.

Side thought:
Maybe payload should be left to specific endpoint or rail. 
|rail | value |
|---------|-------|
|onchain|address|
|lightning|invoice|
|sepa|iban|

specs/payment-endpoint-identifier.md
```diff
@@ -0,0 +1,268 @@
...
+
+## 8. Extending the format
+
+New identifiers are introduced simply by using them, following the
+conventions in Section 5. Before minting a new identifier, authors SHOULD
+check whether an existing identifier already covers the case. Duplicate
```
dzdidi, line 216: this would require some kind of "registry" for identifiers

dzdidi (commented):
Please add a reference to this doc from https://github.com/pubky/paykit-rs/blob/master/README.md and https://github.com/pubky/paykit-rs/blob/master/paykit-lib/README.md

specs/payment-endpoint-identifier.md
ben-kaufman: I wanted to just use some unified key so we could avoid either guessing or having a registry, but if we have a registry (even just in the form of a JSON/ document here) with each endpoint and expected payload that solves it too. What do you think?

specs/payment-endpoint-identifier.md
ben-kaufman: Ideally eventually we could add that so there's no confusion, wasn't sure if we should start with it or not though

specs/payment-endpoint-identifier.md
ben-kaufman: Makes sense, updaded to btc.

specs/payment-endpoint-identifier.md
ben-kaufman: Makes sense, updated to btc-bitcoin...

specs/payment-endpoint-identifier.md
dzdidi: I do not think we need it right away. 

Just wanted to point that out because the fact that we will need to have eventually may impact some design choices now

specs/payment-endpoint-identifier.md
dzdidi: Maybe add some sentence here https://github.com/pubky/paykit-rs/pull/35/changes#diff-3166bbff689fd72f5ce41e7277e0f8aa1c89df830883217f9105b239c57cbfb5R160-R162 that says that additional fields can be of any type (string, number, array, object) and it should be enough

ben-kaufman: Updated, looks good now?

dzdidi (approved):

PR #39, opened 2026-05-18
feat: private payment reference

dzdidi:
- add `PaymentReference` as a `UUID-v4`
- add `PrivateMessageKind` for future extensions to differentiate payment methods from receipts delivery
- add `PrivatePaymentsPayload` for private payments payload


paykit-lib/src/lib.rs
```diff
@@ -241,6 +244,103 @@ pub struct SupportedPayments {
...
+        let reference = reference.into();
+        let uuid = uuid::Uuid::parse_str(&reference).map_err(|err| {
+            PaykitError::Validation(format!("payment reference must be a UUID v4 string: {err}"))
+        })?;
+        if uuid.get_version_num() != 4 {
+            return Err(PaykitError::Validation(
+                "payment reference must be a UUID v4 string".into(),
+            ));
+        }
+        Ok(Self(reference))
```
ben-kaufman: `PaymentReference::new()` validates the UUID but stores the caller's original string. `uuid::Uuid::parse_str` accepts multiple spellings for the same UUID, including simple, braced, and URN forms, so equivalent references can compare/hash/serialize differently. Can we store `uuid.to_string()` here after validation so `PaymentReference` is canonical?

paykit-lib/src/lib.rs
ben-kaufman: This doc still describes the old `entries` parameter, but the function now takes a `PrivatePaymentsPayload`. Can we update this section to mention the required `PaymentReference`, the envelope shape, and that callers still pass the complete entries map inside the payload?

ben-kaufman: The docs here still describe returning `SupportedPayments` / an empty map, but the API now returns `Option<PrivatePaymentsPayload>` and uses `None` when no private message is available. This should be updated so SDK users understand the new no-message behavior and where to access `entries`.

paykit-lib/src/lib.rs
```diff
@@ -241,6 +244,103 @@ pub struct SupportedPayments {
...
+#[cfg(feature = "pubky")]
+/// Private Noise message kinds understood by Paykit.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PrivateMessageKind {
+    PrivatePayments,
+}
+
+#[cfg(feature = "pubky")]
+impl PrivateMessageKind {
+    fn as_str(self) -> &'static str {
```
ben-kaufman, line 307: This takes `self`, but currently ignores it and always returns `"paykit.private_payments"`. Since `PrivateMessageKind` was added for future message-type extensions, this should probably be a `match self { PrivateMessageKind::PrivatePayments => ... }` so adding a new variant forces us to define its wire string instead of accidentally serializing it as private payments.

paykit-lib/src/lib.rs
```diff
@@ -644,37 +744,85 @@ fn compute_private_payment_paths(
...
+) -> Result<String> {
+    let entries = entries
         .iter()
         .map(|(k, v)| (k.as_str(), v.as_str()))
         .collect();
-    serde_json::to_string(&map).map_err(|err| PaykitError::InvalidData {
-        context: format!("failed to serialize private payments JSON: {err}"),
+    let wire = PrivatePaymentsWireRef {
+        version: 1,
+        kind: PrivateMessageKind::PrivatePayments.as_str(),
```
ben-kaufman: `PrivatePaymentsPayload` carries `version` and `kind`, but serialization hardcodes both values here. That lets the payload data diverge from what is actually sent. Can we serialize from the payload instead, e.g. `payload.version` and `payload.kind.as_str()`? Alternatively, if these are intentionally fixed protocol constants for now, the fields should probably not be public/settable on `PrivatePaymentsPayload`.

paykit-lib/src/lib.rs
ben-kaufman: Now that private messages have a `kind`, this drain-and-keep-only-latest behavior is only safe while every message in the queue is a private-payments update. Once receipts or other message kinds are added, a newer receipt could cause `get_private_payments` to discard an older payments payload and then fail on unsupported kind. Should future message kinds use separate queues/paths, or should this loop parse/filter by `PrivateMessageKind::PrivatePayments`?

paykit-lib/src/lib.rs
```diff
@@ -742,14 +890,14 @@ where
 /// - Returns [`PaykitError::InvalidData`] if the map cannot be serialized.
 /// - Returns [`PaykitError::Transport`] if `send_message` fails after all
 ///   retry attempts are exhausted.
-#[instrument(skip(link, entries), fields(count = entries.len()))]
+#[instrument(skip(link, payload), fields(reference = %payload.reference, count = payload.entries.len()))]
```
ben-kaufman: Do we want to log the full `PaymentReference` here? It is not secret material, but it does correlate private payment offers with receipt delivery, so it may be better to log a redacted prefix/hash or omit it from default traces.

paykit-lib/src/lib.rs
```diff
@@ -868,8 +1016,10 @@ where
 ///   is not valid UTF-8 or cannot be parsed as a payments JSON map.
 /// - Returns `Err(PaykitError::Transport)` for decryption or I/O failures.
 #[instrument(skip(link))]
-pub async fn get_private_payments(link: &mut EncryptedLink) -> Result<SupportedPayments> {
-    debug!("receiving private payments map");
+pub async fn get_private_payments(
```
ben-kaufman, line 1210: Since receipts are planned to use Noise messages too, should we add a generic private-message receiver/dispatcher before adding more kinds? Right now `get_private_payments` drains all messages and assumes they are private-payments payloads, so `PrivateMessageKind` is only used for validation rather than routing.

ben-kaufman (changes requested):
A few small cleanup items outside the inline comments:

- Some docs still describe the old private-payments API: raw `HashMap` input, `SupportedPayments` output, and “empty map” when there are no messages. These should match the new `PrivatePaymentsPayload` / `Option<PrivatePaymentsPayload>` API.
- A few comments/log messages still say “private payments map”, maybe rename those to payload/envelope.
- The new public types and constructors could use rustdoc, especially `PrivatePaymentsPayload::new` and `PrivateMessageKind::PrivatePayments`.
- Tests should check that `reference` survives a send/receive round trip. It would also be good to cover unsupported `version`, unsupported `kind`, and invalid `reference`.

paykit-lib/src/lib.rs
dzdidi: The `pub struct PaymentReference(String);` , and while `PaymentReference::new(reference: impl Into<String>)`, there is `let reference = reference.into();` right away. So both validation and version check happen on exactly same version (converted to string)  and in the end of the method there is  `Ok(Self(reference))` so basically it would not compile if it would not be String. BUT according to the docs of `Uuid::parse_str` it is better to use `Uuid::try_parse` instead, so thank you for pointing me into that direction

paykit-lib/src/lib.rs
dzdidi: Good catch :+1:  thank you

ben-kaufman (commented):
Thanks, the main issues from the previous review look fixed now.

A few small cleanup items still remain:

- The README still says `SupportedPayments` is returned by `get_private_payments`, but the API now returns `Option<PrivatePaymentsPayload>`.
- A few comments/log messages still say “private payments map” even though this is now an envelope/payload.
- The new public API could still use rustdoc for `PrivatePaymentsPayload::new`, `reference`, `entries`, and `PrivateMessageKind::PrivatePayments`.
- Tests cover the envelope shape and `PaymentReference` validation, but I’d still like to see a send/receive round-trip assert that `reference` survives, plus explicit envelope parse tests for unsupported `version`, unsupported `kind`, and invalid `reference`.
- `PaymentReference::new()` checks the UUID version nibble, but should probably also require the RFC4122 variant so only real UUID-v4 references are accepted.
- `git diff --check` currently flags trailing whitespace in `paykit-lib/README.md` on the private-payments API bullets.

I don’t see a new correctness issue in the latest dispatcher changes.

ben-kaufman (approved):
Looks good, did you generate the bindings or should I regenerate them?

dzdidi:
@ben-kaufman just pushed one more updated, please have a look. Once good please proceed with bindings :pray:

paykit-lib/src/lib.rs
```diff
@@ -960,23 +960,45 @@ async fn receive_private_messages(link: &mut EncryptedLink) -> Result<usize> {
...
-    kind: PrivateMessageKind,
-) -> Option<BufferedPrivateMessage> {
+) -> Result<Option<PrivatePaymentsPayload>> {
     let mut retained = VecDeque::with_capacity(pending.len());
     let mut latest = None;
+    let mut latest_error = None;
+    let mut malformed = 0usize;
 
     while let Some(message) = pending.pop_front() {
-        if message.is_kind(kind) {
```
ben-kaufman: I think this changes the latest-state behavior a bit. If there is a valid private payments payload, then a newer `paykit.private_payments` message that fails to parse, this would return the older valid payload.

Since private payments are latest-state data, I think the newest same-kind message should still win. So if the newest private payments message is invalid, we should return that error, not fall back to older/stale payment data. Ignoring malformed ones seems fine only when there is a newer valid private payments message after them.

dzdidi: yes, you are right. Reverted

ben-kaufman:
Updated the bindings (and did a tiny doc update). I'll merge and make a new release.

PR #40, opened 2026-05-18
feat: receipts for private payments

dzdidi:
Adds receipt issuance, encrypted receipt storage, receipt-access messages, receipt decryption helpers, and FFI exposure for receipt-related APIs. Also updates private message dispatch so supported message kinds are buffered by type, receipt access is handled as FIFO/event-like data, unknown future kinds are dropped, and malformed receipt-access messages do not prevent later valid messages from being returned.                                                                                                                                          
                                                                                                                                                                                              
Includes validation for UUID-v4 payment references, canonical receipt locations, receipt/location binding during decryption, redacted receipt key formatting, updated docs, and tests for receipt flows, malformed messages, and private-payment envelope parsing


:sweat_smile: After this PR splitting `lib.rs` is a must


ovitrif (changes requested):
No-go for now, but I think this is a narrow fix.

Blocking:
- The generated UniFFI API for private payments now takes/returns `FfiPrivatePaymentsPayload`, but the React Native native wrappers still call the old entries-list shape. iOS still calls `paykitSetPrivatePayments(linkId:entries:)` and treats `paykitGetPrivatePayments` as entries in `paykit-react-native/ios/Paykit.swift`; Android does the same in `paykit-react-native/android/src/main/java/com/paykit/PaykitModule.kt`.
- The generated functions are now `paykitGetPrivatePayments(linkId) -> FfiPrivatePaymentsPayload?` and `paykitSetPrivatePayments(linkId, payload)` in the iOS/Android UniFFI bindings, so the RN native package should fail to build or expose an inconsistent API until those shims and JS/docs are updated.

Non-blocking API thought:
- Since this PR is defining the receipt/private-payment model shape, should we add typed optional `internalId` and `externalId` fields instead of making consumers stash those in `metadata`? I would keep `reference` as the protocol UUID/correlation/location identity, and use the new IDs for app/payment-rail identifiers. If `internalId` is meant to remain local-only, it should not go into shared encrypted payloads.

Validation I ran locally:
- `cargo fmt --check`
- `cargo test`
- `cargo doc --no-deps`

dzdidi:
@ovitrif , thanks, I did not touch bindings much because of my dev env in which I can not verify them (at least iOS). Please can you or @ben-kaufman can address :angel: :pray:

paykit-lib/src/lib.rs
```diff
@@ -918,10 +1078,355 @@ fn decode_private_message(
...
+impl Receipt {
+    fn aad_for_location(location: &str) -> String {
+        format!("paykit.receipt.v1:{location}")
+    }
+
+    /// Encrypt this receipt for storage at `location` using `key`.
+    ///
+    /// The location is authenticated as AEAD associated data; callers must use
+    /// the same location when decrypting.
+    pub fn encrypt(&self, key: &ReceiptDecryptionKey, location: &str) -> Result<String> {
```
ben-kaufman: Do we need `location` as a caller-provided parameter here?

The receipt already has the reference, and the canonical receipt location is derived from that reference. Since decrypt later rejects receipts whose reference does not match the authenticated location, accepting an arbitrary location here means callers can create encrypted receipts that Paykit itself will reject.

Could `encrypt` derive the location internally instead? If we need to keep this parameter, I think we should at least validate it before encrypting.

ben-kaufman (commented):
Leaving one code comment from my review pass.

ben-kaufman (commented):
One small docs wording comment.

CHANGELOG.md
```diff
@@ -7,10 +7,33 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
...
+  `get_receipt_access`, and `decrypt_receipt`.
+- FFI receipt records and exports in `paykit-ffi`: `FfiReceiptDraft`,
+  `FfiReceipt`, `FfiReceiptAccess`, `FfiIssuedReceipt`,
+  `paykit_issue_receipt`, `paykit_get_receipt_access`,
+  `paykit_receipt_location`, and `paykit_decrypt_receipt`.
+
 ### Changed
+- Private encrypted messages now distinguish latest-state private-payment
+  envelopes from event-like receipt-access messages. `get_private_payments`
+  returns the newest valid private-payment envelope, while `get_receipt_access`
```
ben-kaufman: Tiny wording mismatch: I think “newest valid” may overstate the current behavior.

`get_private_payments` collapses pending private-payment messages to the latest message by kind, then parses that selected message. If that latest envelope is malformed, it returns `InvalidData` rather than falling back to an older valid envelope.

Maybe say “newest/latest queued private-payment envelope” unless the intended behavior is to scan for the newest valid one.

README.md
```diff
@@ -61,54 +61,54 @@ Paykit allows you to receive payments from anyone who is aware of the payee's pu
...
 #### Flow
 
-1. The payee creates **Supported Payments List, Methods and Endpoints**  
-2. The payee stores all created data under private locations on Paykit Network associated with their key pair  
-3. The payee grants access to created data to whoever possesses access URL  
-4. The payee shares access URL to **Supported Payments List** with the payer in a secure manner
+1. The peers establish an encrypted link with `initiate_encrypted_link` / `accept_encrypted_link` and drive it with `advance_handshake`.
+2. The payee creates a complete private-payments envelope containing a UUID-v4 `PaymentReference` and the currently supported private payment endpoints.
+3. The payee sends the envelope with `set_private_payments`. Pubky Noise handles encryption, derived private storage paths, file slots, and counterparty retrieval.
+4. The payer calls `get_private_payments` on the same encrypted link and receives the newest valid envelope, if one is available.
```
ben-kaufman: Same wording mismatch is still present in a few user-facing docs: `get_private_payments` picks the latest private-payment message first and then parses that selected message. If that latest envelope is malformed, callers get `InvalidData`; Paykit does not fall back to an older valid envelope. Could we update this and the matching mentions in `README.md`, `paykit-ffi/README.md`, and `paykit-react-native/README.md` to say latest/newest queued envelope instead of “newest valid”?

ben-kaufman (commented):
One remaining docs wording follow-up after the latest push.

CHANGELOG.md
```diff
@@ -7,10 +7,36 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
...
+  `paykit_issue_receipt`, `paykit_get_receipt_access`,
+  `paykit_receipt_location`, and `paykit_decrypt_receipt`.
+
 ### Changed
+- Private encrypted messages now distinguish latest-state private-payment
+  envelopes from event-like receipt-access messages. `get_private_payments`
+  collapses pending private-payment messages to the latest message by kind,
+  then parses that selected message. If that latest envelope is malformed,
+  it returns InvalidData rather than falling back to an older valid envelope.
+  While `get_receipt_access` returns all currently available receipt access
```
ben-kaufman: nit:
```suggestion
  it returns InvalidData rather than falling back to an older valid envelope,
  while `get_receipt_access` returns all currently available receipt access
```

ben-kaufman (approved):
LGTM, I'll make the bindings

ovitrif (approved):
Re-reviewed the follow-up on the previous blocking comments. The RN native shims and JS API now use the FfiPrivatePaymentsPayload shape, generated bindings/package metadata were updated for rc9, receipt encryption derives the canonical location, and the docs wording now reflects latest/newest queued private-payment envelopes. Locally verified: cargo fmt --check, cargo test, cargo clippy --all-targets --all-features, cargo doc --no-deps, and paykit-react-native npm run typecheck.

ovitrif (changes requested):
One late blocker after the code re-review: this PR points SwiftPM at v0.1.0-rc9, but that release/tag/asset does not exist yet. I verified gh release list only shows up to v0.1.0-rc8, git ls-remote has no v0.1.0-rc9 tag, and https://github.com/pubky/paykit-rs/releases/download/v0.1.0-rc9/Paykit.xcframework.zip returns 404. Since Package.swift now depends on that checksum/asset, please create the rc9 release with the generated iOS artifact before merge. The earlier RN payload-shape blocker is addressed and local validation is green.

ben-kaufman:
@ovitrif I'm waiting for approval before making the release so that's intended actually (I generally try to wait unless it blocks other work).

ovitrif (approved):
Re-reviewed after Ben clarified the release flow. The missing v0.1.0-rc9 release is intentional until approval/merge, so I’m not treating the temporary 404 from Package.swift as a blocker. Previous code-review concerns remain addressed: RN native shims and JS API use FfiPrivatePaymentsPayload, generated bindings/version metadata were updated for rc9, receipt encryption derives the canonical location, and docs now describe latest/newest queued private-payment envelopes. Locally verified: cargo fmt --check, cargo test, cargo clippy --all-targets --all-features, cargo doc --no-deps, and paykit-react-native npm run typecheck. Please cut/publish rc9 after merge before mobile consumers depend on the new SwiftPM artifact.

PR #43, opened 2026-05-25
pubky as the only transport

dzdidi:
Making pubky the only transport
Based on  | Blocked by https://github.com/pubky/paykit-rs/pull/40


ben-kaufman (commented):
I think it needs to be rebased since it seems some changes that show up here are already merged from #40

ben-kaufman (commented):
The root README still lists `async-trait` as a dependency for async trait definitions, but this PR removes those traits and drops `async-trait` from `paykit-lib`. Could we remove that dependency line from the README too?

paykit-lib/README.md
```diff
@@ -1,20 +1,16 @@
...
 ## Quick Start
 
 Add `paykit-lib` to your `Cargo.toml`.
 
-To use only the generic transport traits without the Pubky adapters:
-
 ```toml
 [dependencies]
-paykit-lib = { version = "x.x.x", default-features = false }
+paykit-lib = { version = "x.x.x" }
```
ben-kaufman, line 13: The quick start only tells users to add `paykit-lib`, but the public API now uses `pubky::PubkySession` and `pubky::PublicStorage`. Since this crate only re-exports `PublicKey`, users still need a direct `pubky` dependency to follow the examples. Could we add `pubky = "0.8"` to this snippet too?

dzdidi:
See https://github.com/pubky/paykit-rs/pull/46

PR #44, opened 2026-05-25
paykit thesaurus

dzdidi:
## Canonical ubiquitous language for Paykit.

It is needed for better communication with AI.

## Before merge

Once content of `thesaurus.md` is agreed upon, update `agents.md` to rely on terminology from it


THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Pubky Routing
+- **Definition**: Paykit's concrete routing/discovery/storage substrate in practice: Pubky public-key addressing, Pkarr discovery, and homeserver storage.
+- **NOT**: A generic network abstraction in concrete Paykit docs.
+- **Synonyms to AVOID**: Routing Network, Paykit Routing Network
+- **Related terms**: Paykit Protocol, Payment List
+
+### Payment List
+- **Definition**: The payee-published or payee-shared list of Payment Endpoints.
+- **NOT**: The payer-side intersection of compatible endpoints.
```
ben-kaufman: I’d avoid defining this against a payer-side “intersection”. That intersection is not really a Paykit protocol object today. The payer can simply choose one compatible endpoint from the payee’s list without deriving or naming a separate intersection list.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+### Endpoint Format
+- **Definition**: The handle or credential format used on a rail, such as BOLT11, BOLT12, P2TR, IBAN, address, or tag; in identifiers this appears as the third segment.
+- **NOT**: The actual payee-owned payload value.
+- **Synonyms to AVOID**: endpoint when it is ambiguous with Payment Endpoint
+- **Related terms**: Payment Endpoint Identifier, Payment Endpoint Payload
+
+## Private Payments and Receipts
+
+### Private Payment Envelope
+- **Definition**: The latest-state private payment disclosure sent to a Known Peer for a specific Payment Reference. It contains the private Payment Endpoints the payee is willing to share for that payment/request. Latest-wins semantics apply per Known Peer; a newer Private Payment Envelope supersedes older envelopes, even when they have different Payment References.
```
ben-kaufman: I wouldn’t start the definition with “latest-state”. The envelope should be defined by its shape/purpose first: a versioned encrypted Paykit message carrying a Payment Reference and a complete private Payment List. Latest-wins is the retrieval/update semantic and can be mentioned after that, as it already is.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+- **Synonyms to AVOID**: peer reference, relationship reference
+- **Related terms**: Private Payment Envelope, Receipt
+
+### Receipt
+- **Definition**: A Paykit receipt for a payment; Paykit receipts are always encrypted.
+- **NOT**: A plain or unencrypted receipt.
+- **Synonyms to AVOID**: plain receipt, unencrypted receipt
+- **Related terms**: Encrypted Receipt, Receipt Access, Payment Reference
+
+### Encrypted Receipt
```
ben-kaufman: If all Paykit receipts are encrypted today, I’d avoid making `Encrypted Receipt` a separate canonical term.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+- **Synonyms to AVOID**: sender when it obscures payment role
+- **Related terms**: Payee, Supported Payment List, Counterparty
+
+### Payee
+- **Definition**: The party receiving value in a payment flow and publishing or sharing Payment Endpoints.
+- **NOT**: Always a Known Peer.
+- **Synonyms to AVOID**: receiver when it obscures payment role
+- **Related terms**: Payer, Payment List, Payment Endpoint
+
+## Legacy Terms
```
ben-kaufman: I think we shouldn't have this section, or if we keep it we should clearly mention those should not be used and will be replaced once refactoring is done.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+- **Definition**: Implementation/distribution detail under Paykit Library bindings.
+- **NOT**: A first-class Paykit architecture component.
+- **Synonyms to AVOID**: React Native Paykit as product/component language
+- **Related terms**: Paykit Library, Language Bindings
+
+## Forbidden Lexicon
+
+These terms must not be used for new Paykit domain/protocol/component names:
+
+- **Paykit SDK** → use **Paykit Library**.
```
ben-kaufman: I don’t think `Paykit SDK` should be globally forbidden. As discussed, it may eventually mean a higher-level integration package that uses Paykit Library and adds convenience flows. Maybe say: “Do not use Paykit SDK to refer to the current Rust library or language bindings.”

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
+# Paykit Thesaurus
+
+> Canonical ubiquitous language for Paykit. Use this file before naming protocol concepts, public APIs, docs sections, files, types, fields, endpoints, events, or components.
+>
+> Domain rule: the terms below win over older README/code wording when creating new names. Legacy names remain documented so existing code can be read and migrated deliberately.
```
ben-kaufman: I don't think we should keep legacy names here, or if we do we should mention they should be removed post refactoring.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+## Bounded Contexts
+
+### Paykit
+- **Definition**: The whole Paykit system/product, including Paykit Protocol, Paykit Library, future Paykit Daemon work, private payments, receipts, and related components.
+- **NOT**: Only the protocol layer or only the Rust library.
+- **Synonyms to AVOID**: Paykit SDK
+- **Related terms**: Paykit Protocol, Paykit Library, Paykit Daemon
+
+### Paykit Protocol
+- **Definition**: The domain rules, data model, and flows for payment discovery and exchange through Pubky Routing, including public Payment Lists, private payments, Payment Endpoint Identifiers, Payment References, Receipts, and Receipt Access.
```
ben-kaufman: Should avoid “public Payment Lists” here. I think `Payment List` should be visibility-neutral: a payee-controlled collection of Payment Endpoints. It can be published publicly through Pubky public storage or shared privately inside a Private Payment Envelope.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+## Core Protocol Terms
+
+### Pubky Routing
+- **Definition**: Paykit's concrete routing/discovery/storage substrate in practice: Pubky public-key addressing, Pkarr discovery, and homeserver storage.
+- **NOT**: A generic network abstraction in concrete Paykit docs.
+- **Synonyms to AVOID**: Routing Network, Paykit Routing Network
+- **Related terms**: Paykit Protocol, Payment List
+
+### Payment List
+- **Definition**: The payee-published or payee-shared list of Payment Endpoints.
```
ben-kaufman: Suggested definition: “A payee-published or payee-shared collection of Payment Endpoints. A Payment List may be publicly discoverable through Pubky public storage or privately shared inside a Private Payment Envelope.”

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+- **NOT**: The payer-side intersection of compatible endpoints.
+- **Synonyms to AVOID**: Payment Method List, Supported Payments List, Supported Payment List when referring to the payee-published list
+- **Related terms**: Payment Endpoint, Supported Payment List, Payee
+
+### Supported Payment List
```
ben-kaufman: I’m not sure `Supported Payment List` belongs as a core protocol term. It sounds like application/daemon selection logic rather than something Paykit stores, sends, or returns. I’d remove this section, or move it to a future selection/application terms section.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+### Payment Endpoint
+- **Definition**: A whole payee-owned entry in a Payment List, consisting of a Payment Endpoint Identifier and a Payment Endpoint Payload.
+- **NOT**: Only the address, invoice, IBAN, offer, or credential payload.
+- **Synonyms to AVOID**: Payment Option, endpoint payload when referring to the whole entry
+- **Related terms**: Payment Endpoint Identifier, Payment Endpoint Payload, Payment List
+
+### Payment Endpoint Identifier
+- **Definition**: The canonical machine-readable identifier for a payment endpoint type, such as `btc-lightning-bolt12` or `eur-sepa-iban`.
+- **NOT**: The full Payment Endpoint or the payload/credential itself.
+- **Synonyms to AVOID**: PaymentEndpointIdentifier, method id, payment method id
```
ben-kaufman: I don’t think `PaymentEndpointIdentifier` should be a synonym to avoid. In docs, “Payment Endpoint Identifier” is right, but if we rename the Rust type away from `MethodId`, `PaymentEndpointIdentifier` is probably the natural code name.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+- **Definition**: The handle or credential format used on a rail, such as BOLT11, BOLT12, P2TR, IBAN, address, or tag; in identifiers this appears as the third segment.
+- **NOT**: The actual payee-owned payload value.
+- **Synonyms to AVOID**: endpoint when it is ambiguous with Payment Endpoint
+- **Related terms**: Payment Endpoint Identifier, Payment Endpoint Payload
+
+## Private Payments and Receipts
+
+### Private Payment Envelope
+- **Definition**: The latest-state private payment disclosure sent to a Known Peer for a specific Payment Reference. It contains the private Payment Endpoints the payee is willing to share for that payment/request. Latest-wins semantics apply per Known Peer; a newer Private Payment Envelope supersedes older envelopes, even when they have different Payment References.
+- **NOT**: A public Payment List, and not merely a private version of the list if the object includes more than endpoints.
```
ben-kaufman: If `Payment List` is visibility-neutral, I’d avoid saying this is “not” a private list. The envelope is not itself only a list because it also has `version`, `kind`, and `reference`, but it does carry a private Payment List in its `entries`.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+- **Related terms**: Encrypted Receipt, Receipt Access, Payment Reference
+
+### Encrypted Receipt
+- **Definition**: Clarifying synonym for Receipt used when emphasizing storage/security; same concept as Receipt because all Paykit receipts are encrypted.
+- **NOT**: A separate receipt subtype.
+- **Synonyms to AVOID**: plain receipt, unencrypted receipt
+- **Related terms**: Receipt, Receipt Access
+
+### Receipt Access
+- **Definition**: The descriptor/capability that lets a counterparty retrieve and decrypt a Receipt.
```
ben-kaufman: Could this definition be more concrete? `Receipt Access` currently means the descriptor sent over the encrypted link that carries the receipt location and decryption key. It is also event-like/FIFO, unlike private payments which are latest-state. We could elaborate on it here.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+- **Synonyms to AVOID**: React Native Paykit as product/component language
+- **Related terms**: Paykit Library, Language Bindings
+
+## Forbidden Lexicon
+
+These terms must not be used for new Paykit domain/protocol/component names:
+
+- **Paykit SDK** → use **Paykit Library**.
+- **Payment Method List** → use **Payment List**.
+- **Supported Payments List** → use **Supported Payment List** only for the payer-side intersection; use **Payment List** for the payee-published list.
```
ben-kaufman: If we remove `Supported Payment List` as a core protocol term, this forbidden-lexicon entry should go too. For the payee-controlled collection, use `Payment List`, and for payer-side choice/filtering, describe it as payment selection logic rather than naming a separate list.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+- **Payment Option** → use **Payment Endpoint** or **Payment Method**, depending on meaning.
+- **Routing Network** / **Paykit Routing Network** → use **Pubky Routing** in concrete Paykit docs.
+- **MethodId** → use **Payment Endpoint Identifier**.
+- **EndpointData** → use **Payment Endpoint Payload**.
+- **SupportedPayments** → use **Payment List** for the payee-published list.
+- **PrivatePaymentsPayload** → use **Private Payment Envelope**.
+- **Private Payment Method List** / **Private Payment List** → use **Private Payment Envelope** when referring to known-peer private payment data.
```
ben-kaufman: I’d soften this. “Private Payment List” may be useful when referring specifically to the private list carried inside an envelope. What we should avoid is calling the whole `Private Payment Envelope` a list, because the envelope also has message metadata and latest-wins semantics.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+- **Payment Endpoint** for only address/invoice/IBAN/etc. data → use **Payment Endpoint Payload**.
+
+## Component Model
```
ben-kaufman, line 214: The component model should probably remove `Supported Payment List` and add current missing concepts like `Encrypted Link`, `Linked Peer`, `Encrypted Link Handshake`, `Receipt Decryption Key`, `Pubky Homeserver`, and `Private Message Kind`. I’d also avoid listing old implementation names here once we decide to rename them.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+- **Synonyms to AVOID**: receipt token, receipt pointer when naming the protocol concept
+- **Related terms**: Receipt, Encrypted Receipt, Payment Reference
+
+### Counterparty
+- **Definition**: The other party in a payment interaction.
+- **NOT**: Necessarily a Known Peer; the counterparty may or may not have an established encrypted link.
+- **Synonyms to AVOID**: peer when link state matters
+- **Related terms**: Known Peer, Payer, Payee
+
+### Known Peer
```
ben-kaufman: I think we need to separate `Known Peer` from the linked/encrypted state. A Known Peer can be a peer we know or intend to connect with, but may or may not currently have an established encrypted link. Maybe the protocol term for a counterparty with an established encrypted link should be `Linked Peer`.

THESAURUS.md
```diff
@@ -0,0 +1,247 @@
...
+- **Synonyms to AVOID**: routing network protocol, Paykit SDK protocol
+- **Related terms**: Payment List, Private Payment Envelope, Payment Endpoint Identifier, Pubky Routing
+
+### Paykit Library
+- **Definition**: The canonical product/component name for the Rust library that implements and exposes Paykit Protocol functionality to applications.
+- **NOT**: A daemon, SDK, or protocol specification.
+- **Synonyms to AVOID**: Paykit SDK, Paykit PDK
+- **Related terms**: Paykit Protocol, Paykit Daemon, Language Bindings
+
+### Paykit Daemon
```
ben-kaufman, line 27: I think we want a Paykit SDK which wallets and payment processors will integrate rather than a daemon.

ben-kaufman:
I think we should add a few missing terms for the private-message/link layer before this becomes the canonical vocabulary:

- **Encrypted Link**: an established pubky-noise channel used to exchange private Paykit application messages.
- **Encrypted Link Handshake**: the setup flow that establishes an Encrypted Link between two peers.
- **Private Application Message**: a versioned JSON message sent over an Encrypted Link and identified by a Private Message Kind.
- **Private Message Kind**: the `kind` discriminator for private Paykit messages, e.g. `paykit.private_payments` or `paykit.receipt_access`.
- **Latest-State Message**: a private message semantic where only the newest valid message of that kind matters.
- **Event Message**: a private Paykit message where every valid message matters and receivers must process messages in send order.
- **Receipt Decryption Key**: the symmetric key needed to decrypt a Receipt; sensitive material carried in Receipt Access.
- **Receipt Location**: the canonical homeserver path where an encrypted Receipt payload is stored.

These feel important for the refactor because the next modules are likely to be organized around Encrypted Links, private-message dispatch, and receipts.

THESAURUS.md
dzdidi: I see your point. This might be needed for refactoring. Will move to a separate file which can be removed in process of refactoring

THESAURUS.md
dzdidi: I think it is good idea to rename `Known Peer` to `Linked Peer` here. Not sure if need `Known Peer` as a peer without established encrypted link as I think it basically falls under the same category of "non-implemented concepts" as  "Supoorted Payment List"

THESAURUS.md
```diff
@@ -0,0 +1,251 @@
...
+- **Related terms**: Private Application Message, Private Message Kind, Event Message, Private Payment Envelope
+
+### Event Message
+- **Definition**: A private Paykit message where every valid message matters and receivers must process messages in send order.
+- **NOT**: A Latest-State Message where newer messages supersede older messages of the same kind.
+- **Synonyms to AVOID**: event-like message when naming the protocol concept
+- **Related terms**: Private Application Message, Private Message Kind, Latest-State Message, Receipt Access
+
+### Private Payment Envelope
+- **Definition**: a versioned encrypted Paykit message carrying a specific Payment Reference and a complete Payment List of a Know Peer. Latest-State Message semantics apply; a newer Private Payment Envelope supersedes older envelopes, even when they have different Payment References.
```
ben-kaufman: Small wording issue here: `Know Peer` should be `Linked Peer`.

THESAURUS.md
```diff
@@ -0,0 +1,251 @@
...
+- **NOT**: Always a Linked Peer.
+- **Synonyms to AVOID**: receiver when it obscures payment role
+- **Related terms**: Payer, Payment List, Payment Endpoint
+
+## Forbidden Lexicon
+
+These terms must not be used for new Paykit domain/protocol/component names:
+
+- **Payment Method List** → use **Payment List**.
+- **Payment List** for the payee-published list.
```
ben-kaufman: This bullet looks inverted. `Payment List` is the term we want for the payee-published or payee-shared list, so I think this line should be removed.

LEGACY_TERMS.md
```diff
@@ -0,0 +1,47 @@
+# Legacy Terms
+
+This terms are used through codebase and documentation. Their future used must be avoided.
```
ben-kaufman:
```suggestion
These terms are used through codebase and documentation. Their future used must be avoided.
```

THESAURUS.md
```diff
@@ -0,0 +1,250 @@
...
+These terms must not be used for new Paykit domain/protocol/component names:
+
+- **Payment Method List** → use **Payment List**.
+- **Payment Option** → use **Payment Endpoint** or **Payment Method**, depending on meaning.
+- **Routing Network** / **Paykit Routing Network** → use **Pubky Routing** in concrete Paykit docs.
+- **MethodId** → use **Payment Endpoint Identifier**.
+- **EndpointData** → use **Payment Endpoint Payload**.
+- **SupportedPayments** → use **Payment List** for the payee-published list.
+- **PrivatePaymentsPayload** → use **Private Payment Envelope**.
+- **Private Payment Method List** / **Private Payment List** → use **Private Payment Envelope** when referring to known-peer private payment data.
```
ben-kaufman:
```suggestion
- **Private Payment List** for the whole private message → use **Private Payment Envelope**. Use **Payment List** when referring only to the endpoint collection carried inside the envelope.
- **Private Payment Method List** → use **Payment List** or **Private Payment Envelope**, depending on whether you mean the endpoint collection or the whole private message.
```

ben-kaufman (approved):

PR #45, opened 2026-05-25
subscriptions v0.2

dzdidi:
Subscriptions specs

### DIFF
--- a/specs/subscriptions.md
+++ b/specs/subscriptions.md
@@ -1,0 +1,590 @@
+# Paykit Subscriptions Initial Spec v0.2
+
+Status: draft / discussion only  
+Date: 2026-05-19  
+Supersedes: `paykit-subscriptions-v0.1.md`
+
+## v0.2 decisions
+
+This version incorporates the first design decisions after v0.1.
+
+1. `PaymentReference` is **per payment attempt**.
+2. `subscription_id` exists separately from `PaymentReference`.
+3. Subscriptions are private-first: public subscriptions may be mentioned as future work, but implementation scope is private-only for now.
+4. v0.2 defines a common model, but specifies push subscriptions first.
+5. The old URL-secret pull design is replaced by `pubky-noise` private messages.
+6. Subscription proposals may be created by either payer or payee.
+7. Schedule format is simple interval + explicit anchors.
+8. Amount format is string decimal + currency code.
+9. Receipts are method-specific proof containers.
+10. Subscription state is exchanged as private messages and indexed by local daemon DBs.
+11. Pause, resume, cancel, and update are event messages.
+12. The Paykit library exposes primitives; the Paykit daemon provides automation.
+
+## Goal
+
+Define the minimal Paykit subscription model for recurring payments between known peers using:
+
+- `pubky-noise` private messages
+- existing Paykit private payment lists
+- stable `subscription_id` values
+- per-attempt `PaymentReference` values
+- future method-specific `PaymentReceipt` proof messages
+
+## Non-goals for v0.2
+
+- No public subscriptions in implementation scope.
+- No public payment receipts.
+- No URL-secret pull subscription model.
+- No payment-method-specific execution logic.
+- No daemon database schema beyond conceptual indexing requirements.
+- No UI/UX spec.
+- No attempt to replace native recurring-payment mechanisms such as SEPA standing orders, BOLT12 recurrence, app-store subscriptions, etc.
+
+## Core principle
+
+Paykit coordinates recurring payment state and payment-method discovery. It does not execute payments itself.
+
+Paykit is responsible for:
+
+- identifying the counterparty
+- exchanging subscription agreement messages
+- discovering current compatible payment methods/endpoints
+- correlating attempts and receipts
+- giving applications/daemons a common state model
+
+Payment execution remains method-specific and application/daemon-specific.
+
+## Roles
+
+A subscription has two economic roles:
+
+- `payer`: the party expected to pay periodically
+- `payee`: the party expected to receive payments
+
+Either party may create the initial proposal.
+
+Common cases:
+
+- Merchant subscription: payee proposes, payer accepts.
+- Recurring donation: payer proposes, payee accepts.
+
+## Identifier model
+
+### subscription_id
+
+`subscription_id` identifies the long-lived subscription agreement.
+
+Rules:
+
+- UUID-v4.
+- Stable for the life of the subscription.
+- Shared by all lifecycle events for the same subscription.
+- Referenced by payment attempts and receipts.
+
+### PaymentReference
+
+`PaymentReference` identifies one concrete payment attempt.
+
+Rules:
+
+- UUID-v4.
+- Created per payment attempt, not per subscription and not per billing period.
+- Referenced by the corresponding method-specific `PaymentReceipt`.
+- Multiple payment attempts may exist for the same subscription and billing period.
+- Retries must use new `PaymentReference` values unless explicitly modeled as the same attempt by a payment-method-specific layer.
+
+Rationale:
+
+A billing period can have multiple attempts: failed Lightning payment, fallback onchain payment, retry with refreshed endpoint, etc. Attempt-level references make receipts precise and avoid overloading one ID across ambiguous retry flows.
+
+## Privacy scope
+
+Subscriptions are private-only for current implementation.
+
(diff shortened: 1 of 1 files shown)


specs/subscriptions.md
```diff
@@ -0,0 +1,590 @@
...
+  "starts_at": "2026-06-01T00:00:00Z",
+  "anchor": "2026-06-01T00:00:00Z",
+  "ends_at": null
+}
+```
+
+Allowed initial intervals:
+
+- `daily`
+- `weekly`
```
ben-kaufman, line 224: Should the schedule support an interval count, e.g. `{ "every": 2, "unit": "weeks" }` for bi-weekly or `{ "every": 3, "unit": "weeks" }` for every 3 weeks? Also, do we want smaller units like `hourly` or `minutely` for short-lived/micropayment subscriptions, or should those be explicitly out of scope for v0.2?

specs/subscriptions.md
```diff
@@ -0,0 +1,590 @@
...
+
+Rules:
+
+- Each method must be a valid Paykit `MethodId`.
+- The payer daemon may still apply local payment-selection policy within this allowed set.
+- If empty, the payer daemon may treat any mutually supported method as allowed, unless later forbidden by policy.
+
+Open detail for later:
+
+- Whether empty `accepted_methods` should be valid or rejected.
```
ben-kaufman, line 256: I’d say empty should mean no method restriction: the payer can choose anything available from the latest private Payment List / Private Payment Envelope, subject to local policy.

specs/subscriptions.md
```diff
@@ -0,0 +1,590 @@
...
+
+```json
+{
+  "version": 1,
+  "kind": "paykit.subscription_proposal",
+  "subscription_id": "550e8400-e29b-41d4-a716-446655440000",
+  "agreement": {
+    "subscription_id": "550e8400-e29b-41d4-a716-446655440000",
+    "payer": "<payer-pubky>",
+    "payee": "<payee-pubky>",
```
ben-kaufman, line 274: Since the encrypted link already identifies both parties, do we need to include both Pubky keys here? Maybe the agreement can carry a role assignment instead, e.g. `proposer_role: "payer" | "payee"` or `is_proposer_payer: true/false`, and each side derives payer/payee from local vs remote identity.

specs/subscriptions.md
```diff
@@ -0,0 +1,590 @@
...
+
+Rules:
+
+- `payment_reference` is generated per attempt.
+- Multiple attempts may exist for the same billing period.
+- Each attempt should result in either a method-specific receipt or a local failure record.
+
+Open detail for later:
+
+- Whether failed attempts are sent to the counterparty or kept local.
```
ben-kaufman, line 416: The only reason I see to send a failed attempt to the counterparty is if it is actionable, e.g. asking the payee to send a refreshed private Payment List. Otherwise failed attempts may be better as local daemon/SDK state, since broadcasting every failed attempt adds noise and leaks execution details.

specs/subscriptions.md
```diff
@@ -0,0 +1,590 @@
...
+
+The canonical state of a subscription is derived from ordered events, not from a latest-state object.
+
+## Event semantics
+
+Subscription lifecycle messages are event-like.
+
+They must not use latest-state semantics.
+
+Receivers must preserve and process all valid messages in order. Typed getters for one message kind must not discard unrelated message kinds.
```
ben-kaufman, line 140: I think we need to define crash/recovery expectations for event-like subscription messages. Preserving order in memory is not enough for long-lived subscriptions: receivers need to persist received events before treating them as handled, snapshot/restore link progress safely, and dedupe replays after restart. Otherwise a crash after reading an event but before indexing it could corrupt derived subscription state.

specs/subscriptions.md
```diff
@@ -0,0 +1,590 @@
...
+
+This matches the existing `EncryptedLink` buffering direction in `paykit-rs`, where unknown private message kinds are preserved for future typed receivers.
+
+## Message envelope conventions
+
+All private Paykit subscription messages are versioned JSON envelopes.
+
+Common fields:
+
+```json
```
ben-kaufman, line 150: Do event-like subscription messages need a stable `event_id` in the common envelope? `subscription_id` identifies the subscription, but not an individual event. An `event_id` would help with retries, dedupe, crash recovery, and local indexing.

specs/subscriptions.md
```diff
@@ -0,0 +1,590 @@
...
+- ordered/event-like retrieval semantics for receipts and lifecycle events
+
+The library should not:
+
+- schedule recurring jobs
+- execute payments
+- persist daemon state
+- decide payment-method-specific policy
+
+### Paykit daemon
```
ben-kaufman, line 518: I wonder if “Paykit daemon” is too specific as the main higher-level concept. Wallets and payment processors may want a Paykit SDK/runtime embedded in their app or backend, while a daemon is only one deployment shape. Maybe this section should say “SDK/runtime responsibilities”, with daemon as an optional packaging.

specs/subscriptions.md
```diff
@@ -0,0 +1,590 @@
...
+- Each attempt should result in either a method-specific receipt or a local failure record.
+
+Open detail for later:
+
+- Whether failed attempts are sent to the counterparty or kept local.
+
+### paykit.payment_receipt
+
+Carries method-specific proof for one payment attempt.
+
```
ben-kaufman, line 421: Can we clarify whether this is a Paykit `Receipt` or a method-specific payment proof? Current Paykit receipts are payee-issued encrypted receipt payloads with `ReceiptAccess`, while this example looks like payer-sent proof data. If it is proof, maybe `paykit.payment_proof` or `PaymentProof` would avoid overloading `receipt`.

specs/subscriptions.md
```diff
@@ -0,0 +1,590 @@
...
+  "version": 1,
+  "kind": "paykit.subscription_update",
+  "subscription_id": "550e8400-e29b-41d4-a716-446655440000",
+  "agreement": { }
+}
+```
+
+Rules:
+
+- Updates are event messages.
```
ben-kaufman, line 338: Could we define what happens while a `subscription_update` is pending? I think the default should be that the current accepted agreement continues until the counterparty accepts the update. If the sender wants payments to stop immediately, that should be explicit, either by sending a pause/cancellation event or by adding an enum like `pending_update_policy: "continue_current" | "pause_current" | "cancel_current"`. Otherwise implementations may disagree about payments due between the update proposal and acceptance.

specs/subscriptions.md
dzdidi: yes. Very good point. For short units I think it is ok to add them to scheme from day 1 but no need to logical support, in case it makes things more complicated. Am I making sense?

specs/subscriptions.md
dzdidi: I am more inclined towards error in this case. Just to avoid need for some extra handling of edge cases

specs/subscriptions.md
dzdidi: you are right. Do you consider persistance on the homeserver or some other location?

specs/subscriptions.md
dzdidi: you are right, it is better. Either bool or enum is good for me

specs/subscriptions.md
dzdidi: Original differentiation between lib and daemon was whether payment state/history is managed internally (daemon) or externally (internally). Happy to use any other wordings

specs/subscriptions.md
dzdidi: Doc was written before `ReceiptAccess`|`Receipt` came to life but conceptually  I did not distinguish between `Proofs` and `Receipts`. Not sure if this separation needed, I mean what exactly may overload receipts?

specs/subscriptions.md
dzdidi: Maybe some esoteric case like sending TX over network with long processing time like international bank transfer might have some usecase for them

dzdidi: Can be added later if/when necessart

specs/subscriptions.md
dzdidi: If I understand correctly, I would rather go for stopping current payments once there is an update (if possible) simply to avoid potential need to settle the difference

specs/subscriptions.md
ben-kaufman: I think proof is a different concept from Paykit Receipt, especially now that Receipt / ReceiptAccess has a specific meaning. We should define a separate object/message for method-specific payment proof and keep receipts separate.

PR #48, opened 2026-05-28
[codex] Use GitHub token for package publish

ben-kaufman:
## What changed
- Use the built-in GitHub Actions token for the Android Gradle publish step instead of the ORG_PACKAGES_TOKEN secret.

## Why
The release publish job only needs to publish GitHub Packages for this repository, and the workflow already grants packages: write. Using github.token avoids failures from an expired org-level PAT.

## Checks
- Parsed .github/workflows/gradle-publish.yml as YAML with Ruby
- git diff --check


dzdidi (approved):
Lets try

PR #49, opened 2026-05-28
Refactor paykit-lib modules

ben-kaufman:
## Summary

- Split the large `paykit-lib/src/lib.rs` implementation into focused domain modules while keeping the crate root as the public facade/re-export surface.
- Move inline tests into focused module test files for encrypted links, private payment envelopes, public payment endpoints, and receipt access.
- Bump the workspace/binding version to `0.1.0-rc12` and regenerate all platform bindings/artifacts through the build scripts.

## Notes

- The Rust refactor is intended to preserve behavior and public API shape while making the library easier to navigate.

## Verification

- `cargo fmt`
- `cargo doc --no-deps`
- `cargo test -- --test-threads=1`
- `git diff --check`


paykit-lib/src/receipt.rs
```diff
@@ -0,0 +1,159 @@
+mod access;
```
dzdidi, line 1: Shouldn't this be `receipt/mod.rs` ?

paykit-lib/src/tests/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+            _testnet: testnet,
+            sender_link,
+            sender_session,
+            receiver_link,
+            receiver_session,
+        }
+    }
+}
+
+// ── Private Payment Envelopes tests ──────────────────────────────────────────
```
dzdidi: why not in `/test/private_payment_envelopes.rs` ?

paykit-lib/src/tests/mod.rs
```diff
@@ -0,0 +1,256 @@
+use std::collections::HashMap;
```
dzdidi, line 1: Maybe add readme to this folder describing when tests should be here and when should go to the same file where logic is.

My understanding is that e2e tests for to `src/test` and unit tests for to files with logic. 

I am not a big fan of having `src/tests` nesting. Maybe better to move them all the way up to repo root into `/paykit-e2e`  or to `/paykit-lib/test` on the same level as `/paykit-lib/src`

paykit-lib/src/tests/private_payment_envelope.rs
```diff
@@ -0,0 +1,517 @@
+use super::*;
+
```
dzdidi, line 2: Not sure if this is intentional (if so intention is notlear) but some file have pluralized form in naming comparing and some not. For example:
- `test/private_payment_envelopes.rs` - `src/private_payment_envelope.rs`
- test/public_payment_endpoints.rs` - `src/public_payment_endpoint,rs`
- `test/encrypted_link.rs` - `src/encrypted_link.rs`

paykit-lib/src/encrypted_link.rs
```diff
@@ -0,0 +1,17 @@
+mod handshake;
```
dzdidi, line 1: Also looks like `encrypted_link/mod.rs`

dzdidi (changes requested):
also I think there are some references in `AGENTS.md` and probably some other docs which were not updated

paykit-lib/src/receipt.rs
ben-kaufman: I think receipt.rs is fine here. In Rust 2018+ this is the normal style for a module root with submodules under receipt/; receipt/mod.rs is the alternate older-compatible style. The Rust Reference explicitly says the newer convention is encouraged because it avoids many mod.rs files: https://doc.rust-lang.org/stable/reference/items/modules.html#module-source-filenames. So I’d keep this unless we want mod.rs as a repo-wide preference.

paykit-lib/src/encrypted_link.rs
ben-kaufman: Same as https://github.com/pubky/paykit-rs/pull/49#discussion_r3324403875

paykit-lib/src/tests/mod.rs
ben-kaufman: Moved the private-envelope-only fixture into private_payment_envelopes.rs; the remaining helpers in mod.rs are shared by multiple test modules.

paykit-lib/src/tests/mod.rs
ben-kaufman: Added paykit-lib/src/tests/README.md explaining that these are integration-style crate tests that need shared embedded testnet setup or crate-private/test-only helpers, while small pure unit tests should stay next to the logic.

paykit-lib/src/tests/private_payment_envelope.rs
ben-kaufman: The plural names are intentional-ish: source files are named after the implementation module/domain concept, while test files group multiple cases. I agree it is not perfectly uniform, but I don’t think it is worth renaming in this refactor unless we want a strict repo convention.

ben-kaufman:
Updated AGENTS.md and the endpoint identifier spec links. AGENTS.md now describes src/lib.rs as the re-export facade and points private envelope helpers to private_payment_envelope.rs, and the spec now links PaymentEndpointIdentifier to payment_endpoint.rs.

paykit-lib/src/receipt.rs
dzdidi: Thanks for sharing was not familiar

paykit-lib/src/tests/private_payment_envelope.rs
dzdidi: Ok,  then I do not understand why some test files are singular 😅

ben-kaufman: Ok yes fair, I renamed all to singular.

dzdidi (approved):

PR #50, opened 2026-05-29
[codex] Reframe subscriptions as payment requests

ben-kaufman:
## Summary

- Reframes the subscriptions draft as a broader Payment Requests spec.
- Defines one-time Payment Requests and recurring Payment Requests, with subscriptions treated as accepted recurring requests.
- Makes Payment Requests payee-initiated in v0.2 and removes `proposer_role`; payer-initiated recurring payments are treated as wallet/runtime scheduling outside this protocol scope.
- Replaces `subscription_id` terminology with `payment_request_id` and removes v0.2 `paykit.payment_attempt` messages.
- Clarifies expiry, RFC3339 UTC timestamps, month-end recurrence behavior, acceptance by `accepted_event_id`, and Payment Proof semantics.
- Adds Payment Request vocabulary to `THESAURUS.md`, including Payment Request ID, Event ID, Recurrence, Billing Period, Proposal Expiry, Payment Proof, and Payment Endpoint Refresh Request.

## Validation

- `git diff --check docs/subscriptions...HEAD`
- stale-term scan for proposer role, old subscription/payment receipt/method-id/daemon/acceptance-hash wording


specs/payment-requests.md
```diff
@@ -0,0 +1,731 @@
...
+
+This version reframes the previous subscription draft around Payment Requests.
+
+1. A **Payment Request** is the base protocol object: one party requests a payment from the other.
+2. A subscription is an accepted **recurring Payment Request**, not a separate protocol family.
+3. `payment_request_id` identifies the long-lived request or recurring agreement.
+4. `PaymentReference` identifies one concrete payment attempt.
+5. Payment Requests are private-first: public Payment Requests may be explored later, but implementation scope is private-only for now.
+6. The old URL-secret pull design is replaced by `pubky-noise` Private Application Messages.
+7. Payment Request messages may be created by either payer or payee.
```
dzdidi: Why would `payer` create a payment request?

specs/payment-requests.md
```diff
@@ -0,0 +1,731 @@
...
+8. The proposer role is carried explicitly so payer/payee can be derived from the Encrypted Link participants.
+9. One-time requests use `recurrence: null`; recurring requests use an explicit recurrence object.
+10. `expires_at` is required on proposals, but may be `null` for requests with no protocol-level expiry.
+11. Event messages carry stable `event_id` values for dedupe, indexing, and recovery.
+12. Failed payment attempts are local state in v0.2 unless a future event becomes actionable.
+13. Method-specific payment evidence is modeled as Payment Proof, not as Paykit Receipt.
```
dzdidi: I'll need some more info on this. Not 100% sure if I understand what does it mean. So payment-proof is something that comes from payment network? Like tx id + confirmation number for onchain?

specs/payment-requests.md
dzdidi: ?

THESAURUS.md
```diff
@@ -131,10 +137,10 @@
 - **Related terms**: Linked Peer, Private Application Message, Latest-State Message, Payment Endpoint, Payment Reference
 
 ### Payment Reference
-- **Definition**: A per-payment or per-request correlation identifier. In private payment flows, the Private Payment Envelope carries the Payment Reference because the envelope represents the latest private payment disclosure for that payment/request.
-- **NOT**: A stable relationship identifier for a known peer.
-- **Synonyms to AVOID**: peer reference, relationship reference
-- **Related terms**: Private Payment Envelope, Receipt
+- **Definition**: A per-payment-attempt or one-off payment correlation identifier. In Payment Request flows, a new Payment Reference is created for each concrete payment attempt. In private payment flows, the Private Payment Envelope carries the Payment Reference for the concrete payment interaction that needs receiving details.
```
dzdidi: `A per-payment-attempt` reads to me like: if payment with method A failed and trying to payment with method B (sorry for use of retired terminology) then it should NOT be a new reference. I guess it comes from vagueness of what we treat as payment-attempt

specs/payment-requests.md
```diff
@@ -0,0 +1,731 @@
...
+## Core principle
+
+Paykit coordinates payment request state and payment endpoint discovery. It does not execute payments itself.
+
+Paykit is responsible for:
+
+- identifying the counterparty through the Encrypted Link
+- exchanging Payment Request lifecycle messages
+- discovering current compatible Payment Endpoints
+- correlating attempts and proofs
```
dzdidi: If I understand what proofs are than this seem be to a bit contradicitve because to paykit is not deeply aware of the payment method details while payment proofs seem to be very payment specific, no?

specs/payment-requests.md
```diff
@@ -0,0 +1,731 @@
...
+
+- `payer`: the party expected to pay
+- `payee`: the party expected to receive payment
+
+Either role may create the initial proposal.
+
+Common cases:
+
+- Invoice/request for payment: payee proposes, payer accepts or pays.
+- Recurring donation: payer proposes, payee accepts.
```
dzdidi: Why would payee not accept?
I think it is perfectly fine if he does not even know if such request exists

specs/payment-requests.md
```diff
@@ -0,0 +1,731 @@
...
+- For a one-time request, it identifies the single requested payment lifecycle.
+- For a recurring request, it identifies the recurring payment relationship.
+
+### event_id
+
+`event_id` identifies one lifecycle event message.
+
+Rules:
+
+- UUID-v4.
```
dzdidi: Maybe it is better to use some short form so that it is easier to understand what id is where in logs and other places

specs/payment-requests.md
```diff
@@ -0,0 +1,731 @@
...
+- Required on every event-like Payment Request message.
+- Stable across retries/resends of the same event.
+- Reusing the same `event_id` with a different payload is invalid.
```
dzdidi: Maybe event_id should be hash of the event?

specs/payment-requests.md
```diff
@@ -0,0 +1,731 @@
...
+- Used for idempotent storage, replay dedupe, local indexing, and recovery.
+
+`payment_request_id` identifies the request; `event_id` identifies one event within that request. Both are needed.
+
+### PaymentReference
+
+`PaymentReference` identifies one concrete payment attempt.
```
dzdidi: I think we need to have explicit definition of a payment attempt. See https://github.com/pubky/paykit-rs/pull/50/changes#r3332656556

specs/payment-requests.md
```diff
@@ -0,0 +1,731 @@
...
+
+Rules:
+
+- UUID-v4.
+- Created per payment attempt, not per Payment Request and not per billing period.
+- Referenced by the corresponding method-specific `PaymentProof`.
+- Multiple payment attempts may exist for the same Payment Request and billing period.
```
dzdidi: Not sure I understand or agree on this. Let's say lightning payment fails, and user tries on-chain. Then IMO it should be the same reference (different event)

specs/payment-requests.md
```diff
@@ -0,0 +1,412 @@
...
+- `payee`: the party asking to receive payment
+- `payer`: the party being asked to pay
+
+In v0.2, Payment Requests are payee-initiated:
+
+- `paykit.payment_request` MUST be sent by the payee.
+- `paykit.payment_request_acceptance` MUST be sent by the payer.
+- `paykit.payment_request_rejection` MUST be sent by the payer.
+- `paykit.payment_proof` MUST be sent by the payer.
+- `paykit.payment_request_cancellation` MAY be sent by either payer or payee.
```
dzdidi, line 53: I think it is worth adding `paykit.payment_receipt` here

specs/payment-request-protocol.md
```diff
@@ -0,0 +1,412 @@
...
+## Common envelope
+
+Every Payment Request protocol message uses a versioned JSON envelope:
+
+```json
+{
+  "version": 1,
+  "kind": "paykit.payment_request",
+  "event_id": "650e8400-e29b-41d4-a716-446655440000",
+  "payment_request_id": "550e8400-e29b-41d4-a716-446655440000"
```
dzdidi: I suggest to use different ID formats here to make it more meaning of the value more apparent. Maybe UUID and UUID short?

dzdidi: Or to use event hash for event id?

dzdidi: Maybe it is also worth to include (at least optional) payment reference here.

specs/payment-requests.md
```diff
@@ -0,0 +1,412 @@
...
+Rules:
+
+- It is stable for the lifetime of the request.
+- All lifecycle messages for the same request use the same `payment_request_id`.
+- It is not a Payment Reference, Event ID, relationship ID, or billing period ID.
+- A new request requires a new `payment_request_id`.
+
+### Event ID
+
+`event_id` identifies one Event Message.
```
dzdidi, line 113: I am not sure what event is exactly

specs/payment-request-protocol.md
```diff
@@ -0,0 +1,412 @@
...
+
+Rules:
+
+- It is used for idempotent storage and replay dedupe.
+- It is stable across retries of the same event.
+- Each distinct lifecycle message uses a distinct `event_id`.
+
+### Payment Reference
+
+`payment_reference` identifies one concrete payment execution for correlation with Payment Proofs and related artifacts.
```
dzdidi: IMO, if one payment is attempted via Lightning and failed but then retried with onchain or even the same lightning they should share the same reference. The logic is that payment reference is intended to be used for payee side correlation of payment to invoice. Let's say some business invoice always has a number (for internal accounting purposes) then it is good to use this invoice number as a payment reference

specs/payment-requests.md
```diff
@@ -0,0 +1,412 @@
...
+Rules:
+
+- `amount` is required.
+- `amount.value` MUST be a decimal string.
+- `amount.asset` MUST be a currency or asset code.
+- The exact asset registry is out of scope for v0.2.
+- `expires_at` is required and MUST be either `null` or an RFC3339 UTC timestamp using the `Z` suffix.
+- `recurrence` MUST be `null` for one-time requests.
+- `recurrence` MUST be an object for recurring requests.
+- `accepted_payment_endpoint_identifiers` MUST be a non-empty array of valid Payment Endpoint Identifiers.
```
dzdidi, line 166: I have not considered this property before. I assume this is needed to signal support of payment proof verification by payee, right?

specs/payment-requests.md
```diff
@@ -0,0 +1,412 @@
...
+State transitions:
+
+```text
+payment_request -> proposed
+proposed + acceptance -> accepted
+proposed + rejection -> rejected
+proposed|accepted + cancellation -> cancelled
+accepted one-time request + payment_proof -> completed
+proposal with non-null expires_at past expiry -> expired
+```
```
dzdidi, line 377: I think it would not hurt to add terminal states explicitly

specs/payment-request-protocol.md
```diff
@@ -0,0 +1,412 @@
...
+
+## Payment Endpoint lookup
+
+When executing a payment, the payer chooses one of the accepted Payment Endpoint Identifiers and fetches current Payment Endpoint details.
+
+For public endpoints, the payer may use Pubky Routing public storage.
+
+For private endpoints, the payee may have shared a Private Payment Envelope.
+
+The Payment Reference carried by an existing Private Payment Envelope is independent from the `payment_reference` used in a Payment Proof unless a future protocol version defines a stronger relationship. Payment Request payment execution SHOULD create its own Payment Reference for the proof.
```
dzdidi: Not sure if I understand this. Does it imply that
Let's say A and B have private paykit payment channel set up and both exchanged envelopes.
B sends payment request to A.
A accepts it.

A should wait for B to create a new payment evelope?



To me it sounds like we need to have a separate channel for it.
So now we have one channel with kind of like two lanes:
- private payments details (latest wins)
- payment receipts (fifo)
It seems to be worth adding a separate channel with four queues:
- request (latest wins?)
- details (latest wins?)
- proofs (fifo?)
- receipt (fifo?)

specs/payment-requests.md
ben-kaufman: I didn't add it since it already exists so it's not new scope, but yes I'd say we can add that payee may share a receipt (wouldn't say must since it's not really needed for regular payments between contacts for example ie. like in Bitkit)

specs/payment-request-protocol.md
ben-kaufman: Failed attempts are something internal, payment execution just refers here to the successful execution of a payment, doesn't mean a failed attempt needs a different reference. I'll make it a bit more clear.

specs/payment-requests.md
ben-kaufman: Well this comes from the original proposal actually: https://github.com/pubky/paykit-rs/pull/45/changes#diff-7768459de1dd826d3edc02079562ec6bd0d4b13e59e60abe8ab4efcebb459a5eR244

Yes that's the main goal of it.

specs/payment-requests.md
ben-kaufman: Here an Event Message means one individual lifecycle message over the Encrypted Link, for example request, acceptance, rejection, cancellation, or proof.

specs/payment-request-protocol.md
ben-kaufman: Yeah agree the example can be clearer, I’ll make the sample IDs more visually different.

For now I think event_id should stay a UUID. If we make it a hash we need canonical JSON/signing rules, which feels like more scope for v0.2.

I also think payment_reference should stay only where payment execution/proof is involved, not in the common envelope.

specs/payment-requests.md
ben-kaufman: Agree, I’ll add this explicitly.

I think rejected, cancelled, and expired are terminal. completed is terminal for one-time requests. For recurring requests, a proof completes one billing period, not the whole request.

specs/payment-request-protocol.md
ben-kaufman: No, the payer should not need to wait for a new envelope after accepting.

The payer can use current public endpoint details, or a current private envelope if it already has usable details. If it does not have usable details, it cannot execute yet.

I think we can clarify this as logical lanes by message kind: requests/proofs are FIFO event messages, private payment envelopes are latest-state, and receipts are FIFO.

specs/payment-requests.md
dzdidi: Ok, worth adding to thesaurus (IIUC it is Event ID only there)

specs/payment-request-protocol.md
dzdidi: If however payer does not wait will not it cause some conflicts with payment ref? 
Clarification is definitely needed. Specifically this sentence:
> The Payment Reference carried by an existing Private Payment Envelope is independent from the `payment_reference` used in a Payment Proof unless a future protocol version defines a stronger relationship.

specs/payment-requests.md
ben-kaufman: Good point. It was already there as `Event Message`, but the definition was too vague. I updated it here: [THESAURUS.md#event-message](https://github.com/pubky/paykit-rs/blob/codex/payment-requests-spec/THESAURUS.md#event-message) to say it is one FIFO private Paykit message, like Receipt Access or Payment Request lifecycle messages.

specs/payment-requests.md
```diff
@@ -0,0 +1,441 @@
...
+### Event ID
+
+`event_id` identifies one Event Message.
+
+Rules:
+
+- It is used for idempotent storage and replay dedupe.
+- It is stable across retries of the same event.
+- Each distinct lifecycle message uses a distinct `event_id`.
+- In v0.2 it is a UUID, not a hash of the event payload. Hash-based IDs would require canonical serialization rules.
```
dzdidi, line 120: Yes, can be either https://datatracker.ietf.org/doc/html/rfc8785 (https://docs.rs/serde_json_canonicalizer/latest/serde_json_canonicalizer/) or it can also be short uuid. Might make maintenance work easier. Up to you. Just mentioning

dzdidi (approved):
LFG! 

IIRC https://github.com/pubky/paykit-rs/pull/50/changes#diff-b81e49b1e0cc4035aa110befbe088e84e1d789ec2207987dc26a80f7ada80d72R273 will require adjusting code to theasaurus

PR #51, opened 2026-06-03
Implement Payment Requests and private message protocol cleanup

ben-kaufman:
## Summary

- Implement Payment Request / Payment Proof protocol types, parsers, send helpers, and stateless proof/request validation.
- Rename Private Payment Envelope to Private Payment List and move private message consumption to the ordered raw Private Application Message stream.
- Replace typed private receive getters with one ordered raw Private Application Message stream plus stateless parsers, so callers can preserve FIFO order across Payment Requests, Receipt Access, and Private Payment Lists.
- Update receipts to use Receipt ID, retryable prepare/store/send steps, event-shaped Receipt Access, and Payment Request correlation fields.
- Update terminology/docs around Payment References, Payment Amounts, Event Messages, Receipt Access, and the future SDK/runtime boundary.
- Regenerate FFI, iOS, Android, and React Native bindings for the new API surface.

## Notes

`paykit-lib` stays stateless. It does not own lifecycle derivation, role enforcement, dedupe history, recurring scheduling, durable routing, or payment execution. Those belong in a future SDK/runtime layer.

The raw private stream change is intentional: all private message kinds share one Encrypted Link stream, and Event Messages need FIFO handling across kinds. `paykit-lib` now exposes the ordered stream plus stateless parsers, while buffering, dedupe, lifecycle state, Latest-State selection, and ergonomic typed getters belong in a future SDK/runtime.

## Verification

- `cargo fmt --check`
- `cargo check -p paykit-lib --tests`
- `cargo check -p paykit-ffi --all-features`
- `cargo doc -p paykit-lib --no-deps`
- `cargo test -p paykit-lib payment_request::wire --lib`
- `cargo test -p paykit-lib payment_request::types --lib`
- `cargo test -p paykit-lib receipt::tests --lib`
- `cargo test -p paykit-lib private_payment_list::tests --lib`
- `npm run typecheck` in `paykit-react-native`
- `cd paykit-ffi && ./build.sh all`


paykit-lib/src/encrypted_link/private_application_message.rs
```diff
@@ -0,0 +1,356 @@
...
+        )));
+    }
+    Ok(())
+}
+
+pub(super) async fn send_private_application_message(
+    encryptor: &mut pubky_noise::PubkyNoiseEncryptor,
+    max_send_retries: u32,
+    plaintext: &[u8],
+    context: &'static str,
```
dzdidi, line 221: Somewhat ignorant questiont: Would it make sense to have context as enum?

paykit-lib/src/payment_request/types.rs
```diff
@@ -0,0 +1,793 @@
+use std::fmt;
```
dzdidi, line 1: I am tempted to suggest splitting this file into multiple smaller files. Maybe group API, types and wires into same file

specs/paykit-sdk.md
```diff
@@ -0,0 +1,628 @@
...
+   entirely app policy?
+3. Which profile/contact paths should be default versus caller-configured?
+4. Should contact payment resolution default to private-first with bounded
+   recovery and public fallback, or should fallback be app-configured?
+5. How much of endpoint reservation should be generalized now versus left as a
+   payment adapter feature until more integrations need it?
+6. How should the SDK limit stored unknown Private Application Messages, so
+   future protocol messages are not lost but local storage cannot grow forever?
+7. How much scheduling should the SDK own for recurring Payment Requests versus
+   delegating to app/runtime schedulers?
```
dzdidi, line 628: FYI: the paykit-sdk will **probably** be extended with locks related logic

dzdidi (approved):
Some nits. But I am OK to keep things as they are

paykit-lib/src/encrypted_link/private_application_message.rs
ben-kaufman: Yeah could work, but here context is only an internal log/error label, not protocol state. I think keeping it as a static string keeps this simpler for now.

paykit-lib/src/payment_request/types.rs
ben-kaufman: Yeah I also considered that. For now I kept the module split as api/types/wire so public protocol types stay together and wire structs stay separate. We can split further if this grows more, but I’d avoid extra changes like these in this PR.

specs/paykit-sdk.md
ben-kaufman: Makes sense. I think that should fit naturally in the future SDK/runtime layer. I’d leave the current SDK doc generic until we define the locks flow more concretely.

PR #55, opened 2026-06-12
Add Paykit SDK runtime

ben-kaufman:
## Summary

Implements the Paykit SDK layer described in `specs/paykit-sdk.md`: a durable, crash-tolerant runtime above the stateless `paykit-lib` protocol crate.

The SDK gives wallets and apps local state, recovery behavior, retries, backup/restore, and contact payment workflows without each app reimplementing Pubky-backed Paykit logic.

`paykit-lib` remains the low-level stateless protocol and Pubky storage layer.

### New crate: `paykit-sdk`

- **Runtime facade**: `PaykitSdk<S, K, P, C>` for one local Pubky identity, composed from a `StorageAdapter`, `PubkySessionProvider`, `PaymentAdapter`, and `Clock`.
- **Storage contract**: atomic `StorageTransaction` boundary with monotonic ids, per-peer leases, checked updates, and an in-memory test implementation.
- **Identity handling**: Pubky capability tracking, explicit destructive `sign_out`, and session/identity consistency checks.
- **Public Payment Endpoint sync**: publication records, retry/error state, and managed/full namespace cleanup.
- **Encrypted Links**: initiate/accept/advance flows with durable snapshots, generation guards, and per-peer leases.
- **Private stream intake**: ordered private message receive that persists raw items, derived indexes, Event Message dedupe, and the advanced link snapshot in one transaction.
- **Outbound private queue**: FIFO per counterparty, retries with stable Event IDs/payloads, latest-state supersede for Private Payment Lists, and recovery handling for non-retryable link failures.
- **Recovery markers**: minimal public markers for Encrypted Link recovery coordination.
- **Private Payment Lists**: latest-valid-view derivation and publication from adapter receiving details.
- **Endpoint reservations**: contact-scoped receiving details stored atomically with queued private lists.
- **Payment Requests**: propose/accept/reject/cancel/proof APIs plus lifecycle derivation and filterable listings.
- **Receipts**: receipt issuance, Receipt Access delivery/indexing, retrieval, decryption, recipient checks, and conflict handling.
- **Contacts and profiles**: local Contact Records, optional public contact markers, Paykit Profile publish/fetch, Pubky-profile fallback resolution, and blob helpers.
- **Contact payment resolution**: private-first endpoint resolution with optional public fallback and explicit private-state reporting.
- **Backup/restore**: versioned export/restore with identity, capability, snapshot, cross-reference, and counter validation.

### `paykit-lib` additions

- Encrypted Link Recovery Marker helpers.
- Private receive retry classification for deterministic link failures vs retryable transport errors.
- Cursor-paginated Pubky directory listing for complete public Payment Endpoint reads.
- Receipt / Payment Request wire and documentation polish.

### Docs

- Expands `specs/paykit-sdk.md` into the SDK architecture spec.
- Adds `specs/paykit-sdk-bindings.md` for the planned platform binding layer.
- Updates crate READMEs for the SDK/lib boundary.

## Protocol Impacts

- Adds public storage paths for Encrypted Link Recovery Markers under `/pub/paykit/v0/encrypted-link-recovery`.
- Adds SDK profile/contact namespace conventions under `/pub/<namespace>/...`, defaulting to `paykit`.
- Public Payment Endpoints remain under `/pub/paykit/v0/`.
- Latest-State vs Event Message semantics are enforced at the SDK layer.

No changes to the existing Payment Endpoint, Private Payment List, Payment Request, or Receipt wire formats from the already-merged protocol work.

## Bindings Impact

This PR adds the Rust SDK surface.

Existing mobile bindings continue to expose lower-level `paykit-lib` APIs. Platform SDK bindings are the next task and should wrap this SDK as the primary mobile/app integration surface.

## How to Review

Good entry points:

- `paykit-sdk/src/lib.rs`: public SDK surface
- `paykit-sdk/src/storage.rs`: storage adapter contract
- `paykit-sdk/src/runtime/private_stream.rs`: receive/checkpoint flow
- `paykit-sdk/src/runtime/outbound_private.rs`: outbound queue worker
- `paykit-sdk/src/backup/validation.rs`: restore validation
- `specs/paykit-sdk.md`: architecture and behavioral contract

## Verification

- `cargo fmt --check`
- `cargo test`
- `cargo doc --no-deps`

Latest local verification passed across `paykit-lib`, `paykit-sdk`, `paykit-ffi`, and doc tests.

## Follow-ups

- Add platform SDK bindings on top of the Rust SDK surface.
- Add a durable storage adapter conformance suite or reference adapter.
- Add more SDK integration tests with real Pubky session/link flows.
- Tighten provider identity-change handling and multi-device checkpoint behavior.


ovitrif (changes requested):
Thanks for the big SDK lift. I did not find a concrete Rust logic issue in the storage/runtime paths I reviewed, but I need to block on the release/bindings hygiene before this lands.

This PR changes `paykit-lib` public/protocol behavior and exports while leaving the crate/package version and mobile artifact metadata at the already-published `0.1.0-rc12`. `v0.1.0-rc12` already exists, so after merge Android/iOS consumers would not have a new artifact tag/version that contains these `paykit-lib` changes. Please bump the crate/package version, regenerate all bindings with `cd paykit-ffi && ./build.sh all`, update the generated iOS/Android artifacts/package metadata, and create the matching release.

Local verification I ran: `cargo fmt --check`, `cargo test --no-run`, `cargo test`, `cargo doc --no-deps`, and `cargo clippy --all-targets --all-features`.

PS. me writing: this was regular review, I have some issues with cx finding the skill in some sessions, will ask again for review using loop skill once I resolve that

paykit-sdk/src/runtime/outbound_private.rs
```diff
@@ -0,0 +1,446 @@
...
+                                    peer.state,
+                                    LinkedPeerState::Linking
+                                        | LinkedPeerState::RecoveryRequired
+                                        | LinkedPeerState::Blocked
+                                )
+                            })
+                        {
+                            return None;
+                        }
+                        outbound_private_queue_head_is_claimable(
```
ovitrif: At this queue-head filter, `process_pending_private_messages()` only discovers counterparties with claimable `outbound_private_messages`. Terminal private-list messages are skipped by `outbound_private_queue_head_is_claimable()`, but `process_outbound_private_messages(counterparty)` is also the only path that retries `cancel_terminal_private_list_reservations()` when there are no queued messages left. If adapter cancellation fails after `cancellation_started_at` is set, the reservation can stay claimed indefinitely unless the caller already knows to invoke the per-counterparty method. I confirmed this with a temporary repro where an `Invalid` private-list reservation made `process_pending_private_messages()` return no reports. Could we include terminal reservation-cleanup candidates in this pending scan and cover cleanup-only work?

specs/paykit-sdk.md
```diff
@@ -1,628 +1,1289 @@
...
-This is illustrative only.
+`StorageAdapter` is required. It must support atomic updates or an equivalent
+crash-recovery contract.
 
 ```rust
-struct PaykitSdk {
-    // storage, Pubky session access, clock, and payment adapter hooks
+#[async_trait]
+pub trait StorageAdapter {
+    async fn transaction<T>(
```
ovitrif: This snippet still shows adapter authors implementing `transaction<T>`, but the actual `StorageAdapter` requires `transaction_erased`; `transaction<T>` is only the default helper for `Sized` adapters. An implementation written from this spec would not satisfy the trait. Could we update the example to show `transaction_erased` and the boxed callback shape?

specs/paykit-sdk.md
```diff
@@ -1,628 +1,1289 @@
...
+
+Conflicting reused Event IDs must fail closed for the affected derived state.
+
+### PrivatePaymentListView
+
+Latest-state view per counterparty:
+
+- latest valid stream item id
+- current Payment Endpoint map
+- stale/recovery status
```
ovitrif: `PrivatePaymentListView` does not currently expose stale or recovery status; it only has `latest_stream_item_id`, `payment_endpoints`, and `last_refresh_at`. Could we drop this bullet or point readers to the actual recovery/status surface?

specs/paykit-sdk.md
```diff
@@ -1,628 +1,1289 @@
...
+- proposer/payee counterparty
+- current local role: payer or payee
+- immutable terms
+- proposal event id
+- proposal expiry state
+- lifecycle state
+- accepted/rejected/canceled event ids
+- latest Payment Proof records
+- recurrence schedule metadata
+- automatic execution status
```
ovitrif: This section still documents `automatic execution status` and a `paused` lifecycle state, but `PaymentRequestRecord` has no automatic-execution field and `PaymentRequestLifecycleState` has no `Paused` variant. Could we remove these from the current SDK spec or mark them clearly as future work so bindings do not model fields the Rust API never emits?

specs/paykit-sdk.md
```diff
@@ -1,628 +1,1289 @@
...
+## Error Model
+
+SDK errors should be structured:
+
+- `Storage`: durable storage failure
+- `Identity`: Pubky session/key/capability failure
+- `Transport`: Pubky or Encrypted Link transport failure
+- `Protocol`: invalid Paykit message, conflict, or unsupported version
+- `Policy`: operation blocked by configuration or privacy policy
+- `PaymentAdapter`: payment adapter failure
```
ovitrif, line 1234: The error model omits `NotFound`, but `PaykitSdkError` has a distinct `NotFound` variant. Could we include it here so platform mappings preserve missing-resource errors instead of folding them into another category?

ovitrif (commented):
Leaving a few neutral follow-up comments from the review pass.

PR #67, opened 2026-06-12
Add paykit-lib support hooks for SDK runtime

ben-kaufman:
Stack PR 1/6 replacing the earlier SDK review stack. This isolates the low-level paykit-lib support needed by the SDK: Encrypted Link Recovery Marker helpers, private receive retry classification, paginated Payment Endpoint reads, and small protocol/doc refinements. Final stack remains byte-for-byte equivalent to draft PR #55.

### DIFF
--- a/paykit-lib/src/encrypted_link/private_application_message.rs
+++ b/paykit-lib/src/encrypted_link/private_application_message.rs
@@ -6,6 +6,10 @@ use crate::{
     PaykitError, Result,
 };
 
+// Local marker used when decrypted private plaintext is not valid UTF-8.
+// This is not a protocol message. It lets receive callers persist the malformed
+// stream item, advance their local link checkpoint, and avoid wedging on the
+// same encrypted slot forever.
 const INVALID_UTF8_PRIVATE_MESSAGE_PREFIX: &str = "paykit.invalid_utf8_private_message:";
 
 /// Private Message Kind values understood by Paykit.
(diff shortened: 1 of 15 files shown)


paykit-lib/src/encrypted_link/private_application_message.rs
```diff
@@ -1,6 +1,12 @@
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine as _};
 use tracing::{debug, warn};
 
-use crate::{PaykitError, Result};
+use crate::{
+    error::{NonRetryablePrivateReceiveError, NonRetryablePrivateSendError},
+    PaykitError, Result,
+};
+
+const INVALID_UTF8_PRIVATE_MESSAGE_PREFIX: &str = "paykit.invalid_utf8_private_message:";
```
dzdidi, line 13: Not sure I understand the purpose of this together with ` pub fn invalid_utf8_error`

paykit-lib/src/encrypted_link/private_application_message.rs
```diff
@@ -130,6 +136,13 @@ impl PrivateApplicationMessage {
                     .and_then(PrivateMessageKind::parse)
             })
     }
+
+    /// Return a parse error when this item represents non-UTF-8 plaintext bytes.
+    pub fn invalid_utf8_error(&self) -> Option<&'static str> {
+        self.raw_json
+            .starts_with(INVALID_UTF8_PRIVATE_MESSAGE_PREFIX)
+            .then_some("Private Application Message plaintext is not valid UTF-8")
+    }
```
dzdidi, line 154: As mentioned above. Is it only for testing?

dzdidi (commented):
Can you please add some docs on encrypted_link_recovery and market?  I can guess its purpose from the title but some explicit docs on mechanics of its work and usage will be helpful

paykit-lib/src/encrypted_link/private_application_message.rs
ben-kaufman: This is not for testing. The idea is that if a decrypted private message is not valid UTF-8, we still need to return something to the SDK so it can persist the malformed item and advance the local link checkpoint. Otherwise one bad encrypted slot can keep failing receive forever and block later messages. I agree it is not clear enough from the code, so I’ll add a short comment/doc explaining that this marker is local malformed-input handling, not a protocol message.

paykit-lib/src/encrypted_link/private_application_message.rs
ben-kaufman: Same point as above. This is used by the SDK to recognize the local invalid-UTF-8 marker and store a parse error for that stream item. It is not only for tests. I’ll clarify the method docs so it’s clear why this exists and how callers should treat it.

dzdidi (approved):

PR #74, opened 2026-06-17
Add SDK Pubky session bootstrap helpers

ben-kaufman:
## Summary
- add `PubkySessionBootstrap` for SDK-level Pubky signup, signin, session import, and auth handoff helpers
- add Pubky local secret/public key helpers, seed derivation with an app/runtime label, and required capability validation
- wire runtime/session docs so Paykit SDK can own common Pubky session bootstrap flows without requiring Ring or Bitkit-specific code
- update SDK and bindings plans to describe the new generic bootstrap surface

This intentionally keeps Homegate/free-account provisioning out of scope for now. The SDK exposes generic signup with `homeserver_public_key` + optional signup code; apps can still fetch those from Homegate or another provisioning service themselves.

## Validation
- `cargo fmt --check`
- `cargo test -q -p paykit-sdk identity::tests`
- `cargo test -q -p paykit-sdk`
- `cargo clippy -q -p paykit-sdk --all-targets --all-features`
- `cargo doc -q -p paykit-sdk --no-deps`
- `git diff --check`
- `git diff --cached --check`

### DIFF
--- a/paykit-sdk/src/config.rs
+++ b/paykit-sdk/src/config.rs
@@ -127,17 +127,14 @@ impl PaykitSdkConfig {
 
     /// Return the Pubky session capabilities required by this SDK configuration.
     pub fn required_session_capabilities(&self) -> String {
-        let mut capabilities = vec![
-            PAYKIT_SESSION_CAPABILITIES.to_string(),
-            format!("{}:rw", self.paykit_profile_path()),
-            format!("{}:rw", self.paykit_profile_blob_path_prefix()),
-        ];
-
-        if self.public_contact_sharing == PublicContactSharingPolicy::ConfiguredPublicNamespace {
-            capabilities.push(format!("{}:rw", self.public_contact_path_prefix()));
+        if self.profile_namespace == DEFAULT_PROFILE_NAMESPACE {
+            return PAYKIT_SESSION_CAPABILITIES.to_string();
         }
 
-        capabilities.join(",")
+        format!(
+            "{},/pub/{}:rw",
+            PAYKIT_SESSION_CAPABILITIES, self.profile_namespace
+        )
     }
 }
 
--- a/paykit-sdk/src/config.rs
+++ b/paykit-sdk/src/config.rs
@@ -1,7 +1,7 @@
 use serde::{Deserialize, Serialize};
 use std::time::Duration;
 
-use crate::identity::PubkyPublicKey;
+use crate::{identity::PubkyPublicKey, pubky_session::PAYKIT_SESSION_CAPABILITIES};
 
 /// Default namespace segment for SDK profile/contact public data.
 ///
@@ -123,6 +123,21 @@ impl PaykitSdkConfig {
             self.public_contact_path_prefix(),
             public_key.as_str()
         )
+    }
+
+    /// Return the Pubky session capabilities required by this SDK configuration.
+    pub fn required_session_capabilities(&self) -> String {
+        let mut capabilities = vec![
+            PAYKIT_SESSION_CAPABILITIES.to_string(),
+            format!("{}:rw", self.paykit_profile_path()),
+            format!("{}:rw", self.paykit_profile_blob_path_prefix()),
+        ];
+
+        if self.public_contact_sharing == PublicContactSharingPolicy::ConfiguredPublicNamespace {
+            capabilities.push(format!("{}:rw", self.public_contact_path_prefix()));
+        }
+
+        capabilities.join(",")
     }
 }
 
--- a/paykit-sdk/src/config/tests.rs
+++ b/paykit-sdk/src/config/tests.rs
@@ -34,6 +34,10 @@ fn test_config_builds_default_profile_namespace_paths() {
         config.public_contact_path(&public_key),
         format!("/pub/paykit/contacts/{}.json", public_key.as_str())
     );
+    assert_eq!(
+        config.required_session_capabilities(),
+        "/pub/paykit/v0/:rw,/pub/paykit/profile.json:rw,/pub/paykit/blobs/:rw"
+    );
 }
 
 #[test]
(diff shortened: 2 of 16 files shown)


paykit-sdk/src/config/tests.rs
```diff
@@ -34,6 +34,10 @@ fn test_config_builds_default_profile_namespace_paths() {
         config.public_contact_path(&public_key),
         format!("/pub/paykit/contacts/{}.json", public_key.as_str())
     );
+    assert_eq!(
+        config.required_session_capabilities(),
+        "/pub/paykit/v0/:rw,/pub/paykit/profile.json:rw,/pub/paykit/blobs/:rw"
```
dzdidi: just do 
```suggestion
   "/pub/paykit/:rw"
```

paykit-sdk/src/config/tests.rs
```diff
@@ -53,6 +57,23 @@ fn test_config_allows_custom_profile_namespace_segment() {
         config.public_contact_path_prefix(),
         "/pub/bitkit.to/contacts/"
     );
+    assert_eq!(
+        config.required_session_capabilities(),
+        "/pub/paykit/v0/:rw,/pub/bitkit.to/profile.json:rw,/pub/bitkit.to/blobs/:rw"
```
dzdidi: same

paykit-sdk/src/config/tests.rs
```diff
@@ -53,6 +57,23 @@ fn test_config_allows_custom_profile_namespace_segment() {
...
+#[test]
+fn test_config_adds_contact_capability_when_public_contact_sharing_enabled() {
+    let config = PaykitSdkConfig {
+        public_contact_sharing: PublicContactSharingPolicy::ConfiguredPublicNamespace,
+        ..PaykitSdkConfig::default()
+    };
+
+    assert_eq!(
+        config.required_session_capabilities(),
+        "/pub/paykit/v0/:rw,/pub/paykit/profile.json:rw,/pub/paykit/blobs/:rw,/pub/paykit/contacts/:rw"
```
dzdidi: and here

dzdidi: also maybe there is a value in versioning blobs and cotacts too. Not sure

paykit-sdk/src/identity.rs
```diff
@@ -96,6 +103,62 @@ impl PubkyLocalSecretKey {
...
+        })?;
+        Ok(Self::new(bytes))
+    }
+
+    /// Derive a local Pubky secret key from a 64-byte wallet seed.
+    ///
+    /// `runtime_label` must be stable and app/runtime-specific, such as a
+    /// product namespace. Different labels derive different Pubky keys from the
+    /// same wallet seed.
+    pub fn derive_from_seed(seed: &[u8], runtime_label: &str) -> crate::Result<Self> {
```
dzdidi, line 128: I do not remember runtime lable to be documented anywhere where is readmes. I think it should

paykit-sdk/src/pubky_session.rs
```diff
@@ -0,0 +1,497 @@
...
+use url::Url;
+use zeroize::Zeroize;
+
+use crate::{
+    identity::{PubkyIdentityCapability, PubkyLocalSecretKey, PubkyPublicKey},
+    PaykitSdkError, Result,
+};
+
+/// Default Pubky capabilities needed for core Paykit public storage writes.
+pub const PAYKIT_SESSION_CAPABILITIES: &str = "/pub/paykit/v0/:rw";
```
dzdidi: maybe
```suggestion
pub const PAYKIT_SESSION_CAPABILITIES: &str = "/pub/paykit/:rw";
```

paykit-sdk/src/config.rs
```diff
@@ -124,6 +124,21 @@ impl PaykitSdkConfig {
...
+            format!("{}:rw", self.paykit_profile_path()),
+            format!("{}:rw", self.paykit_profile_blob_path_prefix()),
+        ];
+
+        if self.public_contact_sharing == PublicContactSharingPolicy::ConfiguredPublicNamespace {
+            capabilities.push(format!("{}:rw", self.public_contact_path_prefix()));
+        }
+
+        capabilities.join(",")
+    }
```
dzdidi, line 138: I think just `/pub/paykit/:rw` is OK. Happy to be corrected

paykit-sdk/src/config/tests.rs
ben-kaufman: Fixed in `92f6591`. The default required capability is now just `/pub/paykit/:rw`.

paykit-sdk/src/config/tests.rs
ben-kaufman: Fixed in `92f6591`. Custom namespaces now add one namespace capability, so this is `/pub/paykit/:rw,/pub/bitkit.to:rw`.

paykit-sdk/src/identity.rs
ben-kaufman: Fixed in `92f6591`. Added README/spec docs explaining that `runtime_label` should be stable and app/runtime-specific, and changing it derives a different Paykit runtime identity.

paykit-sdk/src/pubky_session.rs
ben-kaufman: Fixed in `92f6591`. `PAYKIT_SESSION_CAPABILITIES` is now `/pub/paykit/:rw`.

paykit-sdk/src/config.rs
ben-kaufman: Fixed in `92f6591`. `required_session_capabilities()` now uses `/pub/paykit/:rw` by default, and adds `/pub/<namespace>:rw` only when the profile namespace is custom.

paykit-sdk/src/config/tests.rs
ben-kaufman: Fixed the capability part in `92f6591`.

On versioning, yeah, there may be value in path versioning later, but I’d keep it out of this PR for now. These are SDK profile/contact/blob paths rather than core protocol routes, and the documents themselves already carry `version`/`kind`. Keeping the path stable lets apps fetch the known profile/contact location and inspect the payload version. If we later need a different namespace layout, we can add path versioning then.

dzdidi (approved):

PR #75, opened 2026-06-17
Organize SDK modules by ownership

ben-kaufman:
## Summary
- move SDK-facing domain modules under `paykit-sdk/src/domain/`
- turn `runtime.rs`, `storage.rs`, and `backup.rs` into folder `mod.rs` files
- keep the public crate re-exports unchanged from `paykit-sdk/src/lib.rs`
- update `specs/paykit-sdk.md` to document the grouped crate layout

This is intended as a structure-only follow-up to the SDK stack. The goal is to make ownership clearer without changing SDK behavior:

- `runtime/` contains workflow methods
- `domain/` contains SDK-facing types, records, and pure helpers
- `storage/` contains storage contracts, queue helpers, and storage records
- `backup/` contains backup/export/restore validation

## Validation
- `cargo fmt --check`
- `cargo test -q -p paykit-sdk`
- `cargo clippy -q -p paykit-sdk --all-targets --all-features`
- `cargo doc -q -p paykit-sdk --no-deps`
- `git diff --check`
- `git diff --cached --check`


dzdidi:
it is a bit wiered to have `/src/*/test.rs` and tests only. Maybe logic should be moved to the same folders?

ben-kaufman:
Yeah makes sense, updated

dzdidi (approved):

PR #82, opened 2026-06-22
Polish SDK mobile bindings

ben-kaufman:
## Summary
- add pure Swift/Kotlin Pubky public-key formatting helpers for app code and unit tests
- expose profile/blob, source-specific payment resolution, private list clearing, and contact private-list sync helpers through SDK FFI
- regenerate iOS/Android FFI outputs and update binding docs

## Notes
- This is stacked on top of the SDK receipt bindings PR.
- React Native wrapper changes are intentionally out of scope for this PR.
- `Package.swift` keeps the release URL/checksum shape; only the Swift source list is updated to include the manual helper.

## Verification
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p paykit-sdk -p paykit-ffi --all-targets`
- `cargo clippy -p paykit-sdk -p paykit-ffi --all-targets --all-features`
- `cargo doc -p paykit-sdk -p paykit-ffi --no-deps`


paykit-sdk/src/runtime/encrypted_links.rs
```diff
@@ -39,6 +39,45 @@ where
...
+            ))),
+            Some(LinkedPeerState::RecoveryRequired) => Err(PaykitSdkError::RecoveryRequired(
+                format!("Encrypted Link recovery is required for counterparty {counterparty}"),
+            )),
+            Some(LinkedPeerState::Blocked) => Err(PaykitSdkError::Policy(format!(
+                "counterparty {counterparty} is blocked"
+            ))),
+            _ => Err(PaykitSdkError::RecoveryRequired(format!(
+                "no active or in-progress Encrypted Link state for counterparty {counterparty}"
+            ))),
```
dzdidi, line 82: maybe split it into firstly checking for Some vs None and then for matching the value?

paykit-sdk/src/runtime/payment_resolution.rs
```diff
@@ -287,6 +432,55 @@ fn private_candidates(
     candidates
 }
 
+pub(super) fn prefer_private_endpoints(resolution: &mut ContactPaymentResolution) {
+    resolution.payable_endpoints.sort_by_key(|endpoint| {
+        if endpoint.endpoint.source == PaymentEndpointSource::PrivatePaymentList {
+            0
+        } else {
+            1
+        }
```
dzdidi: I think this can be simplified

paykit-sdk/src/runtime/private_lists.rs
```diff
@@ -51,6 +52,425 @@ where
...
+            if let Some(error) = counterparty_report.error {
+                report
+                    .failed_to_deliver
+                    .push(PrivatePaymentListDeliveryFailure {
+                        counterparty: counterparty_report.counterparty,
+                        outbound_message_id: None,
+                        reservation_id: None,
+                        error,
+                    });
+            }
```
dzdidi, line 267: can `counterparty_report` contain both `report` and `error`?

paykit-sdk/README.md
```diff
@@ -169,6 +170,10 @@ acknowledgement. Non-retryable link-state failures still pause the peer for
 recovery. Superseded reservation cleanup failures are reported separately and do
 not block delivery of current outbound messages.
 
+Private Payment List helpers support adapter-reserved receiving details. Apps
+can queue reservation-backed lists directly when they already hold reservation
+metadata, including an empty list to clear a counterparty's private list.
+
```
dzdidi, line 176: Not arguing for this to be added now but what do you think of "fall back strategies", should their handling be part of SDK with priority queue being default unless explicitly disabled or should it be outside of SDK entirele?

dzdidi (commented):
Trust @Jasonvdb with reveiw of ffi's. 
Mostly nits from my end

paykit-ffi/src/errors.rs
```diff
@@ -77,8 +77,11 @@ impl From<PaykitSdkError> for PaykitFfiError {
     }
 }
 
-fn format_context(context: String, _source: Option<anyhow::Error>) -> String {
-    context
+fn format_context(context: String, source: Option<anyhow::Error>) -> String {
+    match source {
+        Some(source) => format!("{context}: {source:#}"),
```
Jasonvdb: This leaks private info into app-facing errors.

Folding the whole anyhow chain into the message (`{source:#}`) means `Transport`/`Storage`/`Identity`/`PaymentAdapter` errors now carry the underlying `pubky::Error`'s `Display`, which includes the full request URL and the server response body. For the Encrypted Link recovery marker (`fetch`/`publish`/`remove_encrypted_link_recovery_marker`, reachable from the exported `observe_encrypted_link_recovery_marker`), a failure surfaces something like `...: error sending request for url (pubky://<counterparty_key>/<derived_private_recovery_path>)` in `PaykitFfiError.context`. The old code dropped the source (`_source`) and kept only the static context.

This side-steps the redaction used elsewhere (`FfiPrivateOperationError`, `redacted_app_key`) and exposes storage paths we treat as sensitive. Payment payloads are safe (pubky-noise errors are dataless); the leak is derived private paths, counterparty keys, and server bodies.

Fix: keep only the static `context` here, or route these ops through the existing `FfiPrivateOperationError` redaction so private paths don't land in `error.context`.

paykit-ffi/src/session.rs
```diff
@@ -466,15 +507,15 @@ impl From<PubkyAuthDetails> for FfiPubkyAuthDetails {
             kind: value.kind.into(),
             capabilities: value.capabilities,
             relay_url: value.relay_url,
-            homeserver_public_key: value.homeserver_public_key.map(|key| key.to_string()),
+            homeserver_public_key: value.homeserver_public_key.map(|key| key.to_app_key()),
```
Jasonvdb, line 510: Inconsistency around the homeserver key.

This emits the key in app-key (`pubky...`) form via `to_app_key()`, but the input side parses it raw-only: `sign_up` (`session.rs:244`) and `start_sign_up_auth` (`session.rs:257`) use `PubkyPublicKey::new(...)`, which only accepts a bare 52-char z32. Every other key input in this PR goes through `parse_public_key` -> `from_raw_or_app_key`, which accepts both forms.

Effect: an app that reads `homeserver_public_key` from `parse_auth_deep_link` (app-key form) and passes it back into `sign_up`/`start_sign_up_auth` gets a validation error on a key that works everywhere else, so the round-trip fails.

Fix: parse both spots with `parse_public_key(homeserver_public_key)?` instead of `PubkyPublicKey::new(...)`. Worth a round-trip test feeding the emitted app-key form back into bootstrap.

paykit-sdk/src/runtime/encrypted_links.rs
ben-kaufman: Fixed in 4eb48e7. I split the missing-peer case out before matching the state, so the match now handles concrete `LinkedPeerState` values.

paykit-sdk/src/runtime/payment_resolution.rs
ben-kaufman: Fixed in 4eb48e7. I simplified this to sort on whether the endpoint source is not `PrivatePaymentList`, which keeps private endpoints first without the manual `0`/`1` branch.

paykit-sdk/src/runtime/private_lists.rs
ben-kaufman: Good question. SDK-produced values are exclusive: either `report` or `error`, never both. I added that to the `OutboundPrivateCounterpartySendReport` docs in 4eb48e7 so the invariant is explicit.

paykit-sdk/README.md
ben-kaufman: I think broader fallback strategy should stay policy/configurable rather than hidden default behavior. The SDK covers the current flow now: private-first/public fallback when requested, and adapter ordering for payable endpoints. Priority/failover across receiver strategies feels like a next-layer config/design, not something I would add in this PR.

paykit-ffi/src/session.rs
ben-kaufman: Fixed in 4eb48e7. `sign_up` and `start_sign_up_auth` now use the same raw-or-app-key parser as the rest of the FFI surface, and I added a regression test for the emitted app-key form.

Jasonvdb (approved):

PR #87, opened 2026-07-02
Align Pubky key derivation with Pubky Core

ben-kaufman:
## Summary

Align Paykit's Pubky key derivation helpers with Pubky Core / Pubky Ring.

- Replace Paykit's runtime-label HMAC helper with Pubky Core-compatible BIP39 helpers.
- Add SDK and FFI helpers for deriving a Pubky local secret key from a 64-byte BIP39 seed or an English BIP39 mnemonic phrase.
- Remove the old `derivePubkySecretKey(seed, runtimeLabel)` binding surface.
- Update SDK/FFI docs and regenerate iOS/Android bindings.
- Bump crates and generated platform package metadata to `0.1.0-rc25`.

## Behavior

The new derivation matches Pubky Core: convert the mnemonic to a BIP39 seed with an empty passphrase, then use the first 32 bytes as the Pubky secret key. App/runtime separation should come from receiver folders, Noise keys, and SDK state, not from changing the Pubky identity derivation label.

## Verification

- `cd paykit-ffi && ./build.sh -r --rc all` for the `0.1.0-rc25` bump and full iOS/Android binding build
- `swift package compute-checksum paykit-ffi/bindings/ios/Paykit.xcframework.zip`
- `cargo fmt --check`
- `cargo test -p paykit-sdk -p paykit-ffi --locked`
- `cargo clippy -p paykit-lib -p paykit-sdk -p paykit-ffi --all-targets --all-features --locked`
- `cargo doc -p paykit-lib -p paykit-sdk -p paykit-ffi --no-deps --locked`
- `git diff --check`


ben-kaufman:
Question for @dzdidi before we cut this release:

This PR changes Paykit's helper to match Pubky Core/Ring's current behavior:

`BIP39 mnemonic -> seed with an empty passphrase -> first 32 bytes as the Pubky secret key`

Should Pubky identity derivation intentionally not support a BIP39 passphrase, or should Pubky Core/Ring add explicit passphrase support and Paykit mirror that?

I think Paykit should match Pubky Core as-is unless we decide this at the Pubky layer first, because otherwise Bitkit and Ring can derive different Pubky identities from the same mnemonic.

paykit-ffi/src/session.rs
```diff
@@ -371,13 +371,21 @@ impl FfiPubkyAuthRequest {
     }
 }
 
-/// Derive a local Pubky secret key from a 64-byte wallet seed.
+/// Derive a local Pubky secret key from a 64-byte BIP39 seed.
 #[uniffi::export]
-pub fn derive_pubky_secret_key(
+pub fn pubky_secret_key_from_bip39_seed(
```
Jasonvdb, line 377: Removing `derivePubkySecretKey` breaks the hand-written React Native wrappers, which this PR doesn't update. `paykit-react-native/android/src/main/java/com/paykit/PaykitModule.kt` (import at line 20, call at line 270) and `paykit-react-native/ios/Paykit.swift:293` still call `derivePubkySecretKey(seed, runtimeLabel)`, and the regenerated rc24 bindings no longer define it, so the next `paykit-react-native/scripts/update-bindings.sh` sync fails to compile on both platforms. The JS API (`paykit-react-native/src/index.tsx:193`) also still exposes the now-dead `runtimeLabel` parameter and has no path to the new mnemonic helper.

Fix: migrate the RN package in this PR. Call `pubkySecretKeyFromBip39Seed` / `pubkySecretKeyFromBip39Mnemonic` in both native wrappers, drop `runtimeLabel` from the bridge and TS API, and update the RN README example (`bitkit.to`).

paykit-sdk/src/identity.rs
```diff
@@ -145,34 +143,38 @@ impl PubkyLocalSecretKey {
...
+                context: format!("invalid BIP39 mnemonic phrase: {err}"),
                 source: None,
             }
         })?;
-        mac.update(PUBKY_DERIVATION_CONTEXT);
-        mac.update(&[0]);
-        mac.update(runtime_label.as_bytes());
-        let bytes: [u8; 32] = mac.finalize().into_bytes().into();
-        Ok(Self::new(bytes))
+        Self::from_bip39_seed(&mnemonic.to_seed(""))
```
Jasonvdb: The 64-byte seed returned by `mnemonic.to_seed("")` is dropped without zeroization, and its first 32 bytes are the secret key verbatim. `PubkyLocalSecretKey` already zeroizes on Drop and in `into_inner`, so this temporary falls below the file's own hygiene bar. The parsed `Mnemonic` has the same gap: bip39 is built without its `zeroize` feature, so its word-index array (enough to reconstruct the full recovery phrase) is also dropped un-wiped.

Fix: bind and wipe the seed (`let mut seed = mnemonic.to_seed(""); let key = Self::from_bip39_seed(&seed); seed.zeroize(); key`), and enable `bip39 = { version = "2.2.0", features = ["zeroize"] }` in paykit-sdk/Cargo.toml so `Mnemonic` gets `ZeroizeOnDrop`. Defense in depth only (PBKDF2 internals still make stack copies), but it keeps secret hygiene consistent.

paykit-ffi/src/session.rs
```diff
@@ -371,13 +371,21 @@ impl FfiPubkyAuthRequest {
...
 ) -> Result<Arc<FfiPubkyLocalSecretKey>, PaykitFfiError> {
-    let key = PubkyLocalSecretKey::derive_from_seed(&seed, &runtime_label)?;
+    let key = PubkyLocalSecretKey::from_bip39_seed(&seed)?;
+    Ok(secret_to_ffi(&key))
+}
+
+/// Derive a local Pubky secret key from a BIP39 English mnemonic phrase.
+#[uniffi::export]
+pub fn pubky_secret_key_from_bip39_mnemonic(
+    mnemonic_phrase: String,
```
Jasonvdb, line 388: `mnemonic_phrase` holds the user's full recovery phrase and is dropped un-wiped after derivation; same for the 64-byte `seed: Vec<u8>` in `pubky_secret_key_from_bip39_seed` above. The foreign-language copies can't be wiped from Rust, but the Rust-owned ones can, and the mnemonic is the highest-value secret crossing this boundary.

Fix: take the parameters as `mut` and zeroize them before returning (or wrap them in `zeroize::Zeroizing`), adding the `zeroize` dep to paykit-ffi. Low severity, defense in depth.

paykit-sdk/src/identity.rs
ben-kaufman: Fixed. `from_bip39_mnemonic` now binds the 64-byte seed, derives the key, and zeroizes the seed before returning. I also enabled the `bip39` `zeroize` feature so the parsed mnemonic data follows the same secret hygiene direction.

paykit-ffi/src/session.rs
ben-kaufman: Fixed. I migrated the React Native package off the removed `derivePubkySecretKey(seed, runtimeLabel)` helper. RN now exposes `pubkyPublicKeyFromBip39Seed(seedBase64)` and `pubkyPublicKeyFromBip39Mnemonic(mnemonicPhrase)`, both native wrappers call the new generated BIP39 helpers, and the README no longer references `runtimeLabel` / `bitkit.to` for derivation.

paykit-ffi/src/session.rs
ben-kaufman: Fixed. The FFI seed `Vec<u8>` and mnemonic `String` are now wrapped in `Zeroizing`, so the Rust-owned copies are wiped on return/error. Foreign-language copies are still outside Rust control, as noted.

Jasonvdb (approved):

PR #88, opened 2026-07-02
Add testnet-backed integration tests for paykit-sdk runtime

Jasonvdb:
## Motivation

The SDK's network-facing runtime modules were nearly untested. Every existing runtime test uses a session provider that returns `None`, so all network entry points short-circuit at "no Pubky session available" and the real publish, handshake, send/receive, and recovery paths never execute. Before this PR: `public_endpoints.rs` 5.3% line coverage, `private_stream.rs` 4.2%, `outbound_private.rs` 19.7%, `encrypted_links.rs` 28.9%, `recovery.rs` 38.7%.

paykit-lib solves this with an embedded-testnet harness and sits at 88%. This PR gives paykit-sdk the same treatment.

## What this adds

- `pubky-testnet` (with `embedded-postgres`) as a paykit-sdk dev-dependency. Already in Cargo.lock via paykit-lib, so zero new resolved crates.
- A new `paykit-sdk/tests/testnet_e2e/` integration-test binary that signs up real users on an ephemeral homeserver through the SDK's own `PubkySessionBootstrap`, wraps the resulting `PubkySessionAccess` in a real session provider, and drives the public API end to end.
- 9 tests across 5 flows, each file with at least one failure path:
  - public endpoint sync: publish + stale removal verified through unauthenticated `PublicStorage`; sync after sign-out fails with `Identity`
  - Encrypted Link Handshake: two-party Noise XX to `Linked` over real homeserver files; advance without stored handshake fails with `RecoveryRequired`
  - Private Payment List: enqueue -> send -> receive -> read roundtrip between linked peers; enqueue without a link fails with `RecoveryRequired`
  - recovery markers: publish/observe/remove roundtrip, including fail-closed blocking of private automation during recovery; publish without private link state fails with `Policy`
  - Paykit profiles: publish/fetch roundtrip; absent profile maps homeserver 404 to `Ok(None)`

## Protocol impacts

None. Test-only change, no `src/**` modifications, no public API changes.

## Coverage impact

Workspace line coverage 63.6% -> 69.5% (cargo llvm-cov). The new suite alone exercises 23% of paykit-sdk's lines. Target modules, measured before -> after on this branch:

| module | before | after |
|---|---|---|
| runtime/recovery.rs | 38.7% | 76.9% |
| runtime/outbound_private.rs | 19.7% | 54.8% |
| runtime/public_endpoints.rs | 5.3% | 52.0% |
| runtime/encrypted_links.rs | 28.9% | 51.1% |
| runtime/private_stream.rs | 4.2% | 38.7% |
| runtime/profiles.rs | 16.6% | 37.6% |
| pubky_session.rs | 35.9% | 50.7% |

## Validation

- `cargo test -p paykit-sdk --test testnet_e2e`: 9 passed, 0 failed, 22.9s
- `cargo test --workspace --all-features`: runs under `cargo llvm-cov` above, all green
- `cargo fmt --all --check`: clean
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: clean

## Follow-ups (out of scope)

- Advanceable test clock for lease-expiry and retry-backoff paths
- Crash-safety/replay tests (persist-order invariants, homeserver write faults)
- Receipts, Payment Requests, contacts, and backup end-to-end flows

### DIFF
--- a/paykit-sdk/tests/testnet_e2e/public_endpoints.rs
+++ b/paykit-sdk/tests/testnet_e2e/public_endpoints.rs
@@ -2,6 +2,14 @@ use paykit_sdk::{load_public_endpoint_records, PaykitSdkError, PublicationStatus
 
 use crate::harness::{build_testnet, receiving_detail, TestUser};
 
+/// Payload published for `identifier`, or `None` when the endpoint is absent.
+fn payload_of<'a>(list: &'a paykit_lib::PaymentList, identifier: &str) -> Option<&'a str> {
+    list.payment_endpoints
+        .iter()
+        .find(|(id, _)| id.as_str() == identifier)
+        .map(|(_, payload)| payload.as_str())
+}
+
 #[tokio::test]
 async fn test_sync_public_endpoints_publishes_and_removes_managed_endpoints() {
     let testnet = build_testnet().await;
@@ -33,14 +41,11 @@ async fn test_sync_public_endpoints_publishes_and_removes_managed_endpoints() {
     let list = paykit_lib::get_payment_list(&storage, &payee)
         .await
         .expect("Payment List fetch should succeed");
-    assert!(list
-        .payment_endpoints
-        .keys()
-        .any(|id| id.as_str() == "btc-lightning-bolt11"));
-    assert!(list
-        .payment_endpoints
-        .keys()
-        .any(|id| id.as_str() == "btc-onchain"));
+    assert_eq!(
+        payload_of(&list, "btc-lightning-bolt11"),
+        Some("lnbc-test-invoice")
+    );
+    assert_eq!(payload_of(&list, "btc-onchain"), Some("bc1q-test-address"));
 
     // Local publication records mirror the remote state.
     let records = load_public_endpoint_records(&user.storage)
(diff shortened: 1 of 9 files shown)


dzdidi (approved):

paykit-sdk/tests/testnet_e2e/public_endpoints.rs
```diff
@@ -0,0 +1,102 @@
...
+    // Remote state: both endpoints readable through unauthenticated storage.
+    let storage = user.access.outbox_client.public_storage();
+    let payee = user
+        .public_key
+        .to_public_key()
+        .expect("public key conversion should succeed");
+    let list = paykit_lib::get_payment_list(&storage, &payee)
+        .await
+        .expect("Payment List fetch should succeed");
+    assert!(list
```
ben-kaufman: This only checks that the endpoint IDs exist, so the test would still pass if `sync_public_endpoints` wrote the wrong payload for either ID. Could we assert the exact payloads from `get_payment_list` before and after removal?

Jasonvdb: Good catch, the keys-only check would indeed miss a wrong or swapped payload. Fixed in 20902fa: added a payload_of helper and the test now asserts the exact payload per identifier after both syncs, including that the surviving endpoint keeps its original payload after the removal.

ben-kaufman (approved):
Looks good after the payload assertion fix in 20902fa.

PR #89, opened 2026-07-02
Scope Paykit storage by receiver path

ben-kaufman:
## Summary

Adds receiver-scoped Paykit storage paths so one Pubky identity can host multiple app/runtime-owned Paykit receivers without sharing private runtime state.

The receiver path shape is exactly `{app}/{wallet|server}`, for example `bitkit/wallet` and `bitkit/server`.

Public Paykit data is now scoped like:

- `/pub/paykit/v0/{receiver_path}/...`

Encrypted Link transport/recovery files are receiver-scoped like:

- `/pub/paykit/v0/private/{receiver_path}/...`

The runtime model remains app-owned: each receiver owns its own local Encrypted Link state, private stream checkpoints, outbound queue, recovery markers, receipts, and endpoint publication.

## What Changed

- Adds validated `PaykitReceiverPath` support in `paykit-lib`.
- Scopes public Payment Endpoint storage, Receipt storage, Receipt Access, and Encrypted Link Recovery Markers by receiver path.
- Scopes Encrypted Link path derivation and snapshots by local and remote receiver paths, and rejects mismatched snapshot restores.
- Updates SDK config/runtime calls to require explicit receiver configuration instead of broad flat Paykit paths.
- Updates FFI/session capability helpers so apps ask for the scopes required by their SDK config.
- Removes flat Paykit path helpers/constants from the active v0 flow.
- Regenerates Swift/Kotlin bindings, bumps crates/Android/Swift package metadata to `0.1.0-rc33`, and refreshes the Swift package checksum.
- Updates docs, thesaurus, and changelog for receiver-path terminology and paths.

## Notes

- Private workflows require explicit local and counterparty receiver paths.
- Generic receiver discovery, receiver indexes, signing, aggregation, and cross-receiver history are follow-up layers.
- Rebased on latest `master` after PR #90; Encrypted Link recovery outbox cleanup is now receiver-scoped too.

## Verification

- `cd paykit-ffi && ./build.sh --release --rc all` passed, producing `0.1.0-rc33` artifacts and publishing `com.synonym:paykit-android:0.1.0-rc33` to Maven Local.
- `cargo fmt --check` passed.
- `git diff --check` passed.
- `cargo test -q -p paykit-sdk --lib` passed.
- `cargo test -q -p paykit-ffi --lib` passed.
- `cargo test -q -p paykit-lib pubky_routing:: --lib` passed.
- `cargo clippy -q -p paykit-lib -p paykit-sdk -p paykit-ffi --all-targets --all-features` passed.
- `cargo doc -q -p paykit-lib -p paykit-sdk -p paykit-ffi --no-deps` passed.
- `git lfs fsck` passed.

Note: `cargo test -q -p paykit-lib encrypted_link --lib` was attempted, but local embedded Postgres startup failed before assertions (`failed to start embedded postgres: Request failed after 3 retries`).


paykit-sdk/src/storage/in_memory.rs
```diff
@@ -99,14 +99,25 @@ impl StorageTransaction for StorageStateTransaction {
...
+        self.state.linked_peers.insert(
+            (
+                record.counterparty.clone(),
+                record.counterparty_receiver_path.clone(),
+            ),
+            record,
+        );
     }
 
     fn contact_records(&self) -> Vec<ContactRecord> {
```
Jasonvdb, line 123: `contact_records()` sorts by `public_key` only, but `state.contact_records` is now keyed by `(PubkyPublicKey, PaykitReceiverPath)`, so one pubkey can hold several records and their relative order is nondeterministic.

The sort at `in_memory.rs:130` leaves same-pubkey records in `HashMap::values()` iteration order, which varies per run, and this reaches callers through the public `PaykitSdk::contact_records()` API. The backup exporter already uses the two-level `(public_key, receiver_path)` sort, so this site was just missed.

Two sibling listings have the same gap:
- `linked_peers()` (`runtime/mod.rs:458`), sorted by `counterparty` only.
- `sync_public_contact_markers` (`runtime/contacts.rs:237`), sorted by status then `public_key` only.

Fix: add a receiver-path tiebreaker to all three, e.g. `.then_with(|| left.receiver_path.as_str().cmp(right.receiver_path.as_str()))` (use `counterparty_receiver_path` for `linked_peers`), matching the tiebreaker already used in `pending_outbound_private_counterparties` and `payment_request_counterparties`.

paykit-sdk/src/runtime/payment_requests.rs
```diff
@@ -67,23 +88,42 @@ where
         }
         let now = self.clock.now();
 
-        let counterparties = if let Some(counterparty) = &filter.counterparty {
-            self.ensure_peer_not_blocked(counterparty).await?;
-            vec![counterparty.clone()]
+        let counterparties = if let (Some(counterparty), Some(receiver_path)) =
```
Jasonvdb: This filter only honors the counterparty scope when both `counterparty` and `counterparty_receiver_path` are `Some`. `PaymentRequestFilter` derives `Default` and `counterparty_receiver_path` defaults to `None`, so a `(Some, None)` filter falls into the enumerate-all `else` branch.

Two consequences vs. the prior behavior:
- The blocked-peer check (`ensure_peer_not_blocked`) only runs in the `Some/Some` arm, so a filter naming a blocked counterparty now returns `Ok(vec![])` instead of `Err(Policy)` (the enumerate-all branch just omits blocked peers). The caller loses the signal that the peer is blocked.
- It derives and recovery-marks records for every counterparty (each a storage transaction), then discards all but one via `filter.matches()`, so a single-counterparty query does O(all-counterparties) work.

Results stay correctly scoped (`filter.matches()` re-applies the counterparty check), so this is a lost error signal plus wasted work, not a data-correctness bug.

Fix: reject the `(Some, None)` shape as a validation error, or scope it to the named counterparty across its receiver paths with the blocked-peer check preserved. No internal caller builds this shape today, so a test pinning it would help.

paykit-lib/src/encrypted_link/link.rs
```diff
@@ -264,15 +284,43 @@ pub async fn close_encrypted_link(mut link: EncryptedLink) -> Result<()> {
...
 pub async fn restore_encrypted_link(
+    session: pubky::PubkySession,
+    secret_key: [u8; 32],
+    remote_pubkey: &PublicKey,
+    local_receiver_path: &PaykitReceiverPath,
+    remote_receiver_path: &PaykitReceiverPath,
+    outbox_client: pubky::Pubky,
+    snapshot: EncryptedLinkSnapshot,
+) -> Result<EncryptedLink> {
+    snapshot.validate_receiver_scope(local_receiver_path, remote_receiver_path)?;
```
Jasonvdb, line 295: This `validate_receiver_scope` check is the production side of the PR's "rejects mismatched snapshot restores" control (the SDK calls `restore_encrypted_link` / `restore_encrypted_link_handshake` with raw params from `runtime/outbound_private.rs:287` and `runtime/private_stream.rs:136`), but nothing tests it. The mismatch tests in `tests/encrypted_link.rs` only exercise the separate `validate_private_payment_paths` check in the `_from_config` variants, which the SDK does not use here.

Same gap for the new `SNAPSHOT_WIRE_VERSION` gate in `deserialize_snapshot_wire` (`snapshot.rs:148`): the test helper `scoped_snapshot_bytes` only ever emits version 1, so the reject-on-version-mismatch path never runs.

Fix: add two failure-path tests, per the repo guideline for new protocol features. Restore with mismatched receiver paths asserting `validate_receiver_scope` rejects, and a version-2 wire asserting the gate returns `InvalidData`.

paykit-sdk/src/storage/queue.rs
```diff
@@ -96,6 +98,7 @@ pub(super) fn supersede_outdated_private_payment_lists(
         .iter()
         .filter(|message| {
             &message.counterparty == counterparty
+                && &message.counterparty_receiver_path == counterparty_receiver_path
```
Jasonvdb, line 101: This `counterparty_receiver_path` filter (and the matching one in `claim_next_outbound_private_message`) is the core of the PR's cross-receiver isolation invariant, but no test exercises the skip path. Every storage/domain test hardcodes the single receiver path `bitkit/wallet`, so nothing asserts that a Private Payment List queued for `(X, r1)` is not superseded by, or claimed under, `(X, r2)`.

The same is true across the re-keyed stores: no test checks that `linked_peer` / `encrypted_link_state` / `payment_endpoint_reservation` / `receipt_record` lookups return `None` for a different receiver path after a save. A future refactor that dropped one of these receiver-path filters (cross-receiver state bleed, the PR's primary threat) would pass the entire suite.

Fix: add at least one positive isolation test that saves state for `(X, r1)` and asserts lookups/claims/supersede for `(X, r2)` do not see it.

paykit-sdk/src/storage/in_memory.rs
ben-kaufman: Thanks. I added the receiver-path tiebreaker to linked_peers in 7be6ac2. The contact_records / public contact marker part is stale now: ContactRecord is keyed by public_key and carries a sorted receiver_paths list, so there are no multiple same-pubkey records to order there; sync_public_contact_markers works over those same records.

paykit-sdk/src/runtime/payment_requests.rs
ben-kaufman: Fixed in 7be6ac2. The counterparty-only filter now scopes to that counterparty across all receiver/runtime paths and still runs the blocked-peer check for each matching receiver path, so the Policy signal is preserved. Added a regression test for both the multi-receiver listing and blocked-peer behavior.

paykit-lib/src/encrypted_link/link.rs
ben-kaufman: Fixed in 7be6ac2. I added failure-path coverage for the raw restore APIs used by the SDK: restore_encrypted_link_handshake and restore_encrypted_link now have mismatched receiver-scope tests, and the snapshot wire-version gate has an unsupported-version test.

paykit-sdk/src/storage/queue.rs
ben-kaufman: Fixed in 7be6ac2. Added a cross-receiver isolation test that saves state for one receiver path and asserts a different receiver path cannot see the linked peer, encrypted link state, reservation, receipt, or queued outbound message; it also verifies queue claiming only claims the matching receiver path.

Jasonvdb (approved):

PR #90, opened 2026-07-03
Fix SDK auth startup and stale link recovery

ben-kaufman:
## Summary
- Fix Pubky Ring auth startup by making SDK/FFI auth start/resume helpers async, so mobile calls run inside UniFFI tokio runtime.
- Let `ensure_link_with_peer` start a fresh deterministic Encrypted Link handshake in the same call after stale stored handshake state fails, whether restore fails or restore succeeds but the handshake cannot advance.
- Preserve an already in-progress recovery handshake when observing a remote recovery marker, so peers with mutual recovery markers do not repeatedly reset relink attempts.
- Clear the local Encrypted Link outbox before fresh recovery handshakes, so peers do not read stale ciphertext left at deterministic private stream paths after delete/recreate flows.
- Keep Paykit release artifacts at unpublished `0.1.0-rc28` and regenerate all iOS/Android bindings/native artifacts.

## Why
Bitkit mobile saw SDK-side failures in the Ring auth and post-delete/recreate recovery flows:
- Ring auth startup could fail before Ring opened with `there is no reactor running`, because auth-flow creation happened from a synchronous FFI method.
- After deleting/recreating the same Pubky identity and re-adding a contact, stale Encrypted Link state could get retried forever. One case failed during restore with `RestoreReplayError`; another restored but failed while advancing with `failed to transition to transport mode: IsHandshake`. Both now mark the peer recovery-required and allow `ensure_link_with_peer` to start a fresh handshake instead of retrying the same stale snapshot.
- In rc27, two peers with recovery markers could still cycle through initiate/accept/advance because marker observation could clear a fresh in-progress recovery handshake. Marker observation now leaves active handshakes alone and lets the relink finish or fail on its own state.
- The final delete/recreate blocker was deterministic private stream storage: the same identity pair derives the same private paths, so old ciphertext can remain at low counters and be read by the peer after a fresh relink. Recovery now clears the local write stream, which is the peer read stream, before starting a fresh recovery handshake.

## API Impact
- Swift/Kotlin `startSignInAuth`, `startSignUpAuth`, and `resumeAuth` are now async/suspend.
- `paykit-lib` now exposes `clear_encrypted_link_outbox` for deleting the local encrypted write stream during recovery.
- `Package.swift` targets `v0.1.0-rc28` with the regenerated XCFramework checksum.
- Android Gradle/Cargo package versions remain `0.1.0-rc28`.

## Verification
- `cd paykit-ffi && ./build.sh --release --rc all`, including iOS XCFramework generation and Android Maven Local publication
- `cargo fmt --check`
- `cargo test -p paykit-sdk --lib recovery --locked`
- `cargo test -p paykit-sdk --lib --locked`
- `cargo test -p paykit-ffi --locked`
- `cargo clippy -p paykit-lib -p paykit-sdk -p paykit-ffi --all-targets --all-features --locked`
- `cargo doc -p paykit-lib -p paykit-sdk -p paykit-ffi --no-deps --locked`
- `cargo test -p paykit-sdk --test testnet_e2e --no-run --locked`
- `swift package compute-checksum paykit-ffi/bindings/ios/Paykit.xcframework.zip`
- Bitkit Android preserved delete/recreate E2E passed with these local artifacts, per mobile integration testing.

Note: the focused live testnet E2E for mutual recovery markers was added and compiles, but this local machine could not start embedded Postgres (`failed to start embedded postgres: Request failed after 3 retries`), so I could not run that test to completion here.


paykit-sdk/src/runtime/recovery.rs
```diff
@@ -370,6 +370,14 @@ where
                 marker_created_at,
             )
             .await?;
+        if changed {
+            paykit_lib::clear_encrypted_link_outbox(
```
Jasonvdb: This clear runs after the peer link lease has been released, so it can race a fresh handshake started by a concurrent `ensure_link_with_peer`. `mark_remote_recovery_marker_observed_if_needed` claims and releases the lease internally, so nothing serializes this delete against other peer link operations. A concurrent `ensure_link_with_peer` can claim the lease, see `RecoveryRequired`, clear the outbox itself, start the fresh handshake, and write handshake message 1 to the deterministic write path; the in-flight clear here then deletes that message. The counterparty never sees message 1, the new handshake wedges in `Pending`, and `recovery_handshake_is_in_progress` suppresses further marker-based recovery for that peer. Trigger: background marker observation running concurrently with a UI-triggered relink for the same counterparty, which is the normal mobile shape.

Fix: perform the clear while the lease is still held, matching the ordering in `start_link_handshake_with_claim` (encrypted_links.rs:628), e.g. claim the lease, mark, clear, then `finish_peer_link_operation`.

paykit-sdk/src/runtime/recovery.rs
```diff
@@ -610,17 +632,30 @@ fn parse_recovery_marker_created_at(marker: &EncryptedLinkRecoveryMarker) -> Res
...
+            marker_created_at.timestamp() <= checkpointed_at.timestamp()
+        });
+    let marker_is_before_private_receive = peer
+        .and_then(|peer| peer.last_private_receive_at)
+        .is_some_and(|received_at| marker_created_at.timestamp() <= received_at.timestamp());
+    marker_is_before_link_checkpoint || marker_is_before_private_receive
+}
+
+fn recovery_handshake_is_in_progress(link_state: Option<&EncryptedLinkStateRecord>) -> bool {
+    link_state.is_some_and(|state| state.handshake_snapshot.is_some())
```
Jasonvdb: This guard has no staleness bound, so any stored `handshake_snapshot` suppresses remote recovery markers indefinitely. A handshake wedged in `Ok(Pending)` never errors (advance keeps returning Pending while the counterparty's next message is missing), so nothing clears the snapshot, and every future marker is ignored at recovery.rs:396 and recovery.rs:451. A counterparty who lost state mid-handshake can then publish markers forever with no effect, and the pair cannot converge. The pre-rc31 behavior (a marker newer than `checkpointed_at` resets local state) had churn but not this deadlock.

Fix: bound the guard, e.g. only skip the marker when the snapshot's `checkpointed_at` is within a freshness window of `marker_created_at`, so an aged Pending snapshot yields to a newer marker. Worth a test where a marker newer than a stale Pending snapshot unwedges the pair.

paykit-sdk/src/runtime/recovery.rs
ben-kaufman: Fixed in `d4ed4ce`. The remote-marker path now claims the peer link operation lease before applying the marker, clears the deterministic outbox while that lease is still held, and only then releases the lease. That keeps a concurrent relink from writing a fresh handshake message that the marker observer can delete afterward.

paykit-sdk/src/runtime/recovery.rs
ben-kaufman: Fixed in `d4ed4ce`. The in-progress handshake guard is now bounded by the peer link operation freshness window: recent handshakes are still preserved, but a newer remote recovery marker can take over once the pending handshake is stale. I also added `test_remote_recovery_marker_observation_accepts_newer_marker_after_stale_handshake` for this case.

Jasonvdb (approved):

PR #94, opened 2026-07-08
Add Paykit receiver discovery markers

ben-kaufman:
## Why

Receiver-scoped Paykit paths let one Pubky identity expose multiple runtimes such as `bitkit/wallet` and `bitkit/server`. Some receivers may intentionally publish no public Payment Endpoints, but still need to be discoverable from a Pubky identity so apps can decide which receiver paths to try or maintain.

This PR adds a small public receiver marker document at `/pub/paykit/v0/{receiver_path}/receiver.json`. The marker contains only the receiver path and coarse capabilities. It does not contain payment details or payment method hints. Receivers that want to remain unlisted can skip publishing it and still be reached through explicit Pubky key + receiver path intents.

## What changed

- Added `PaykitReceiverMarker` and `PaykitReceiverCapabilities` to `paykit-lib`.
- Added publish, fetch, remove, parse, and serialize helpers for receiver markers.
- Updated receiver-path discovery to include receivers with valid markers even when they have no public endpoints.
- Kept discovery strict: malformed markers are invalid data, empty marker files are invalid, and unknown capability fields are rejected.
- Added SDK methods to fetch, publish, and remove the configured receiver marker.
- Added a SDK guard so public-only sessions cannot advertise private receiver capabilities.
- Exposed the marker APIs through FFI and regenerated iOS/Android bindings.
- Updated docs/specs and added unit coverage.

## Verification

- `cargo fmt --check`
- `cargo test -p paykit-lib receiver_marker:: -- --test-threads=1`
- `cargo test -p paykit-sdk --lib -- --test-threads=1`
- `cargo test -p paykit-ffi --lib -- --test-threads=1`
- `cargo clippy -p paykit-lib -p paykit-sdk -p paykit-ffi --all-targets --all-features`
- `cargo doc -p paykit-lib -p paykit-sdk -p paykit-ffi --no-deps`
- `git diff --check`
- `cd paykit-ffi && ./build.sh all`

The Pubky-backed `receiver_marker_round_trip_and_discovery` test was added, but local execution is blocked by embedded Postgres shared-memory exhaustion (`could not create shared memory segment: No space left on device`).


paykit-lib/src/pubky_routing.rs
```diff
@@ -137,7 +196,9 @@ pub async fn fetch_paykit_receiver_paths(
                     continue;
                 }
             };
-            if seen_receiver_paths.insert(receiver_path.clone()) {
+            if seen_receiver_paths.insert(receiver_path.clone())
+                && receiver_is_publicly_advertised(storage, owner, &receiver_path).await?
```
Jasonvdb, line 200: `fetch_paykit_receiver_paths` returns `Err` for the whole identity if any single `receiver.json` is malformed, so one bad marker hides every sibling receiver.

`receiver_is_publicly_advertised` calls `fetch_paykit_receiver_marker`, which returns `PaykitError::InvalidData` (not `Ok(None)`) when a marker is empty, invalid JSON/UTF-8, has an unsupported `version`/`kind`, has unknown fields, or has a mismatched `receiver_path`. The `.await?` on this line propagates that error out of the loop and aborts the entire listing.

Trigger: a newer client publishes `bitkit/server/receiver.json` with `version: 2`, or a partial write leaves an empty `receiver.json`. An older client can then no longer discover `bitkit/wallet` either, even though it has valid public endpoints. This is inconsistent with the loop right above (which `continue`s past malformed receiver-path segments) and with `list_resources`/`fetch_text` degrading 404/empty to `None`/empty list.

Fix: treat a marker parse/`InvalidData` outcome as "not advertised via marker" and fall through to the endpoint check (or `continue`, logging at debug), the same way malformed segments are skipped. Keep propagating genuine `Transport` errors so real network failures still surface.

paykit-lib/src/pubky_routing.rs
```diff
@@ -147,6 +208,26 @@ pub async fn fetch_paykit_receiver_paths(
     Ok(receiver_paths)
 }
 
+async fn receiver_is_publicly_advertised(
+    storage: &PublicStorage,
+    owner: &PublicKey,
+    receiver_path: &PaykitReceiverPath,
+) -> Result<bool> {
+    if fetch_paykit_receiver_marker(storage, owner, receiver_path)
```
Jasonvdb: Minor, not a defect: discovery now makes up to ~2 extra sequential requests per receiver here (a marker GET, then an `endpoints/` LIST on a marker miss). This is the intended marker-vs-endpoint filter and N is small per identity, so it's fine as-is. If discovery latency ever matters, the per-receiver `receiver_is_publicly_advertised` checks are independent and could run concurrently (e.g. `join` over the futures) rather than sequentially.

paykit-lib/src/pubky_routing.rs
ben-kaufman: Good catch. This is already fixed on the current head: discovery now catches `PaykitError::InvalidData` from `receiver.json`, logs it, and falls through to endpoint discovery while still propagating real transport errors. I also added regression coverage for an invalid sibling marker plus a valid endpoint-backed receiver.

paykit-lib/src/pubky_routing.rs
ben-kaufman: Agreed. I’m leaving this sequential for now because the current discovery path is simple and N should stay small per identity. If it shows up in practice we can parallelize these independent checks later.

paykit-lib/src/pubky_routing.rs
Jasonvdb: Resolved in `b0dfba0`: `receiver_is_publicly_advertised` now matches on the marker fetch and treats `InvalidData` as "not advertised via marker", falling through to the endpoint check while still propagating `Transport` errors. `aa15cae` adds `invalid_receiver_marker_does_not_block_sibling_discovery` and `empty_receiver_marker_is_invalid_and_does_not_advertise_receiver`, confirming a corrupt/empty sibling marker no longer hides receivers with valid endpoints. Verified against head `aa15cae`.

Jasonvdb (approved):

PR #95, opened 2026-07-14
Add generic Pubky Auth companion claim support

ben-kaufman:
## Summary

Adds a generic, high-level Pubky Auth companion-claim operation and exposes it through the rc35 Swift and Kotlin bindings.

The integrator supplies only:

- the auth URL query parameter name
- the claim type
- the exact expected capability scope
- the application-serialized unsigned payload

Paykit validates that configuration against the `pubkyauth://` request, binds the payload to the approving Pubky identity and individual auth secret, encrypts and delivers it to the claim-specific relay channel, and only then approves the normal AuthToken.

Bitkit can use this with:

- `x-bitkit-claim`
- `watch-only-account-v1`
- `/pub/paykit/v0/bitkit/server/:rw`
- its own serialized version/index/address-type/78-byte-xpub payload

No Bitkit schema, xpub parser, address enum, or capability constant is embedded in Paykit.

## Protocol impact

- Signature input:
  `UTF8(query_parameter + "|" + claim_type + "|") || SHA256(decoded_auth_secret) || unsigned_payload`
- Signed plaintext:
  `unsigned_payload || ed25519_signature[64]`
- Relay channel:
  `base64url_no_pad(BLAKE3(claim_type || "|" || decoded_auth_secret))`
- Relay body:
  `nonce[24] || XSalsa20-Poly1305 ciphertext/tag`
- Companion delivery completes before normal Pubky Auth approval.
- Errors distinguish invalid auth requests, invalid claim descriptions, encryption failures, relay delivery failures, normal authorization failures, and neutral unexpected failures.

## SDK and bindings

- Adds validated Rust `PubkyAuthCompanionClaim`.
- Exposes `PubkyAuthCompanionClaim` and `approveAuthWithCompanionClaim` in Swift and Kotlin.
- Uses `Data` on Swift and `ByteArray` on Kotlin for the opaque unsigned payload.
- Reuses Pubky's parsed sign-in/signup relay and secret instead of maintaining a second auth URL decoder.
- Keeps only the protocol-required direct runtime dependencies: base64url, BLAKE3, and XSalsa20-Poly1305. The HTTP relay server remains test-only.
- Removes the direct `bs58` and `percent-encoding` dependencies because payload parsing is integrator-owned and auth URL decoding is Pubky-owned.
- Redacts payloads from Rust Debug output, documents the generated-platform logging risk, and sanitizes relay failures so URLs, relay channels, and response bodies do not escape.
- Maps unknown future SDK errors to a neutral append-only FFI `Unexpected` case.
- Bumps and regenerates all checked-in artifacts at `0.1.0-rc35`; rc33 is already released and the public rc34 tag is not reused.
- Replaces the Bitkit-specific protocol document with a generic companion-claim specification and a Bitkit example.

## Verification

- `cargo fmt --check`
- `cargo test --workspace --all-features`
  - paykit-sdk: 366 passed
  - paykit-ffi: 46 passed, 1 intentionally ignored live-server test
  - paykit-lib: 180 passed
  - testnet E2E: 10 passed
  - doc tests: 6 passed, 11 ignored
- `cargo doc --no-deps`
- `cd paykit-ffi && ./build.sh -r --rc all` — iOS and all four Android ABIs built; Swift/Kotlin regenerated; Android Maven-local publish succeeded
- XCFramework checksum and zip integrity verified; Android AAR includes all four ABI libraries.


paykit-sdk/src/pubky_session/companion_claim.rs
```diff
@@ -0,0 +1,330 @@
...
+const MAX_QUERY_PARAMETER_LEN: usize = 64;
+const MAX_CLAIM_TYPE_LEN: usize = 128;
+
+/// App-defined data delivered alongside one Pubky Auth approval.
+///
+/// The integrating application owns the claim's payload schema and supplies
+/// its unsigned binary representation. Paykit owns request validation,
+/// request-bound identity signing, encryption, relay delivery, and approval
+/// ordering. It does not expose the underlying cryptographic primitives.
+#[derive(Clone, Debug, PartialEq, Eq)]
```
Jasonvdb: `PubkyAuthCompanionClaim` derives `Debug`, so any `{:?}` log prints the raw `unsigned_payload` bytes. The spec says plaintext payloads must not appear in normal logs (specs/pubky-auth-companion-claims.md:103), and for Bitkit this payload is an account xpub, which exposes the account's full transaction history. Neighboring sensitive types hand-write redacting Debug impls (`PubkyLocalSecretKey`, `PubkySessionSecret`, `PubkyAuthDetails`), so the derive here is an easy accidental leak. Same issue on `FfiPubkyAuthCompanionClaim` (paykit-ffi/src/session.rs:135), where the generated Kotlin data class `toString()` also prints the payload.

Fix: hand-write `Debug` to redact the payload, e.g. `.field("unsigned_payload", &format_args!("<redacted, {} bytes>", self.unsigned_payload.len()))`, and note in the FFI record docs that apps must not log the record.

paykit-sdk/src/pubky_session/companion_claim.rs
```diff
@@ -0,0 +1,330 @@
...
+            request.relay.clone(),
+            derive_companion_channel_id(claim_type, &request.secret),
+        )
+        .map_err(invalid_auth_url)?;
+        channel
+            .produce(self.pubky.client(), encrypted_claim)
+            .await
+            .map_err(
+                |err| PubkyAuthCompanionClaimApprovalError::RelayDeliveryFailure {
+                    reason: err.to_string(),
```
Jasonvdb: `RelayDeliveryFailure { reason: err.to_string() }` can embed the full request URL on transport failures. reqwest error strings include the URL, which here is the relay plus the BLAKE3-derived channel id, and the reason flows verbatim through the FFI into app logs and crash reporters. The channel id is stable for the auth request, so a log reader could consume or fetch the encrypted claim on a retry. `InvalidAuthUrl` is explicitly documented as redacted; this path should match.

Fix: replace `err.to_string()` with a sanitized reason (error kind or HTTP status, no URL).

paykit-ffi/src/session.rs
```diff
@@ -524,6 +591,31 @@ impl From<PubkyAuthDetails> for FfiPubkyAuthDetails {
...
+            PubkyAuthCompanionClaimApprovalError::EncryptionFailure { reason } => {
+                Self::EncryptionFailure { reason }
+            }
+            PubkyAuthCompanionClaimApprovalError::RelayDeliveryFailure { reason } => {
+                Self::RelayDeliveryFailure { reason }
+            }
+            PubkyAuthCompanionClaimApprovalError::AuthorizationFailure { reason } => {
+                Self::AuthorizationFailure { reason }
+            }
+            other => Self::AuthorizationFailure {
```
Jasonvdb: The wildcard arm maps any future SDK error variant to `AuthorizationFailure`, whose contract is "approval failed after companion delivery succeeded". A future pre-delivery variant would then read as fail-open (claim already delivered) when it may be fail-closed, and this distinction is exactly how integrators are meant to tell the two states apart. The wildcard itself is required (the SDK enum is `#[non_exhaustive]`), but its target can be neutral.

Fix: map unknown variants to a neutral fallback (e.g. `Unexpected { reason: String }`) that makes no delivery-state claim. Needs a bindings regen since it changes the exposed error enum.

paykit-sdk/src/pubky_session/companion_claim.rs
ben-kaufman: Fixed in 2ff082c. Both the SDK and FFI records now use manual Rust `Debug` implementations that redact `unsigned_payload` and show only its byte length, with regression tests. The generated Swift/Kotlin docs also explicitly warn callers not to log or stringify the record. Kotlin's UniFFI-generated data-class `toString()` (and Swift reflection) cannot be customized here without a larger binding API redesign, so the platform risk is documented rather than claimed to be redacted.

paykit-sdk/src/pubky_session/companion_claim.rs
ben-kaufman: Fixed in 2ff082c. Relay failures no longer stringify the underlying request error: server failures retain only the numeric HTTP status, transport failures use a fixed category string, and other failures use a fixed generic string. Regression tests exercise both a canary server response and a closed-port transport failure and assert that the relay URL, derived channel, auth secret, and response body are absent.

paykit-ffi/src/session.rs
ben-kaufman: Fixed in 2ff082c. The wildcard now maps to a new neutral `Unexpected` FFI variant with a fixed non-sensitive reason and no delivery-state claim. It was appended after the existing variants to preserve their UniFFI tags, and the regenerated Swift/Kotlin bindings expose it as tag 7. A regression test verifies that its display text does not imply companion delivery succeeded.

Jasonvdb (approved):

PR #96, opened 2026-07-14
Fix FFI error-mapping diagnostics loss and add tests

Jasonvdb:
## Problem

The `PaykitSdkError -> PaykitFfiError` mapping (`paykit-ffi/src/errors.rs`) had mapping bugs, and only 1 of the 8 arms had a test. Review iterations then showed the deeper problem: error `context` is rendered verbatim into the generated Kotlin/Swift exception message, and several error paths folded sensitive data into it (Pubky request URLs and response bodies via cause chains, DH-derived private Receipt Locations, and serde/decode detail derived from decrypted plaintext and fetched documents).

## Changes

FFI mapping (`paykit-ffi/src/errors.rs`):

- `format_context` no longer folds an SDK error's anyhow `source` chain into `context`. Those sources are dropped entirely, not logged: recovery-marker URLs can embed DH-derived private storage paths, non-2xx response bodies are sensitive, and `tracing::debug!` would ship them to any enabled subscriber. For ordinary SDK errors only the outer `context` label crosses the FFI; app-authored `PaykitFfiError` callback errors are recovered separately by downcast as described below.
- The `PaymentAdapter` arm of `ffi_error_to_sdk` now passes `source` through like the other source-bearing arms, fixing the `FFI -> SDK -> FFI` round-trip downcast.
- The `#[non_exhaustive]` wildcard arm forwards the variant's `Display` with the stable `unsupported_sdk_error` code instead of a constant placeholder.

Receipt Location redaction (`paykit-sdk`, `paykit-lib`):

- `store_encrypted_receipt_error`, `missing_encrypted_receipt_error`, and `store_prepared_receipt_error` are explicit constructors that drop the Receipt Location and keep a static label. The persisted retrieval-record error string keeps the location for local diagnostics only; it never crosses the FFI.
- A missing Encrypted Receipt (confirmed 404/GONE) in `retrieve_receipt` now surfaces as `PaykitSdkError::NotFound` (FFI code `not_found`) instead of `Transport`. `merge_retrieval_error` keeps `NotFound` as the weakest failure so it never masks a transport or decrypt error from another Receipt Location.

Decrypted-plaintext and fetched-document redaction (`paykit-lib`, `paykit-sdk`):

- serde_json errors embed verbatim document fragments on type mismatches. Parse errors for decrypted receipt plaintext (`Receipt::decrypt`) and Receipt Access JSON (`parse_receipt_access_json`, carries the Receipt Decryption Key) now use `invalid_plaintext_json`: static context, no source.
- `From<PaykitError> for PaykitSdkError` no longer folds `InvalidData` sources into the `Protocol` string, which crosses the FFI verbatim. Only the lib-supplied `context` string survives; on the paths redacted in this PR that context is a static label.
- The encrypted-envelope parse and recipient-public-key contexts are static; detail stays in `source`, which no longer crosses.
- The encrypted-envelope base64 and nonce-length decode errors, the fetched-body UTF-8 check, Receipt Decryption Key validation, and backup-restore encrypted-receipt validation also use static labels (base64 `DecodeError` Display names the offending byte and offset; for Receipt Decryption Key material that detail is dropped entirely, not even retained in `source`).
- `validate_wire_version_kind_str` (shared by the Receipt, Receipt Access, and Payment Request event wire parsers; the Private Payment List parser validates version/kind inline and is part of the known follow-up) and the encrypted-envelope version/kind/algorithm check no longer echo the offending values; both use static contexts.

Callback error fidelity (`paykit-ffi`, `paykit-sdk`):

- `PaykitSdkError::NotFound`, `Protocol`, `Policy`, and `RecoveryRequired` are restructured from plain-string tuple variants into struct variants with the same `{ context, source: Option<anyhow::Error> }` shape as the other four, making all eight variants uniform.
- All eight arms of `ffi_error_to_sdk` now stash the original callback error in `source`, so the reverse conversion recovers the exact variant, custom machine-readable code, and reason by downcast. No callback error code degrades to a generic one on the round trip.
- The new `source` fields cannot smuggle network-derived or decrypted-plaintext data across the boundary: the FFI conversion drops every `source` unless it downcasts to a `PaykitFfiError`, and only paykit-ffi's own plumbing stashes one (platform callback errors via `ffi_error_to_sdk` and the payment-adapter wrapper, plus paykit-ffi-authored validation/encoding errors). paykit-lib and paykit-sdk never construct a `PaykitFfiError`. Pinned by a sentinel test per variant.
- `From<paykit_lib::PaykitError>` keeps dropping lib `InvalidData` causes entirely: `PaykitSdkError` derives field-wise `Debug`, so a retained cause would surface in `format!("{err:?}")` and structured Rust logs even though the FFI conversion never forwards it. Only the lib-supplied `context` string survives (a static label on the paths redacted in this PR).
- `CHANGELOG.md` gains an Unreleased upgrade note covering the variant shape change, the `not_found` contract change, the redacted exception text, and the round-trip fix.

Tests:

- Table-driven coverage of all 8 `From<PaykitSdkError>` arms; `test_ffi_sdk_round_trip_recovers_original_via_downcast` pins lossless recovery (variant + custom code + reason) for all eight variants.
- Redaction regression tests with sentinel values: private-path URL and response body in the cause chain, Receipt Location in all three constructors (variant pinned to `NotFound` for retrieval), decrypted-plaintext sentinel through real AEAD encrypt/decrypt, Receipt Access sentinel, the `InvalidData` source kept out of the `Protocol` context string, and a per-variant sentinel test proving the four new `source` fields never reach FFI `context` or `Display`.
- Exact-equality pins on the encrypted-envelope decode messages and Receipt Decryption Key validation messages, so a reintroduced dynamic `: {err}` suffix fails the test.
- SDK sentinel tests pin the static fetched-body UTF-8 and backup-restore encrypted-receipt errors, assert `source: None`, and prove private content reaches neither `Display` nor field-wise `Debug`.
- `merge_retrieval_error` priority tests.

## Protocol / bindings impact

- Protocol impact: none. No UniFFI shape change; no binding regeneration needed.
- Breaking Rust API change to `paykit-sdk`: `PaykitSdkError::NotFound`, `Protocol`, `Policy`, and `RecoveryRequired` change from tuple to struct variants (~350 construction/match sites updated in-repo). Downstream apps consume the generated bindings, whose error shape is unchanged.
- Behavioral contract change for generated clients: a missing Encrypted Receipt during `retrieve_receipt` now throws the `not_found` exception (code `not_found`) instead of `transport_error`. Apps matching on the transport variant for this case must update.
- SDK `source` cause chains, which can contain request URLs and response bodies, are no longer rendered into Swift/Kotlin exception text; app-authored `PaykitFfiError` callback errors are recovered verbatim by design. The listed inbound Receipt, Receipt Access, and Receipt Decryption Key paths use static parse/decode/validation contexts (envelope serde/base64/length, decrypted-plaintext and Receipt Access parses, fetched-body UTF-8, key validation, and backup-restore validation). Payment Request and Private Payment List parse detail remains the known follow-up below. An explicitly named sensitive-diagnostics export can be added in a follow-up if raw causes need to be retrievable.

## Known follow-up

The same decrypted-plaintext detail folds exist in the Payment Request and Private Payment List paths, with different coverage:

- The Payment Request event parsers already route their version/kind checks through the shared redacted helper, so only their serde-level parse detail remains.
- The Private Payment List parser (`parse_private_payment_list_json`) is not covered by that helper: it validates version/kind inline and echoes the offending values, folds serde detail, and echoes invalid Payment Endpoint Identifier keys.
- The SDK's outbound private-message validation (`validate_outbound_private_message`) and the backup validation folds around it have the same class of echoes, and there are assorted non-plaintext dynamic contexts (directory-listing identifiers in `pubky_routing`, Encrypted Link snapshot decode folds).

PR #104 has merged, so the cleanup is unblocked and will follow as a separate PR using the same static-label pattern.

## Verification (local)

All green on a `master`-based checkout:

- `cargo test --workspace`
- `cd paykit-ffi && ./build.sh all` (iOS and Android; Android native symbols/alignment and Maven-local publication validated)


paykit-ffi/src/errors.rs
```diff
@@ -69,16 +69,21 @@ impl From<PaykitSdkError> for PaykitFfiError {
...
     }
 }
 
-fn format_context(context: String, _source: Option<anyhow::Error>) -> String {
-    context
+fn format_context(context: String, source: Option<anyhow::Error>) -> String {
+    match source {
+        // `{:#}` prints the full anyhow cause chain, so diagnostics crossing to
+        // Swift/Kotlin retain the underlying cause instead of only the outer label.
+        Some(source) => format!("{context}: {source:#}"),
```
ben-kaufman: `{source:#}` can include the full Pubky request URL and a non-2xx response body. Recovery marker URLs contain a DH-derived private storage path, and the generated Kotlin exception puts `context` directly in its message. Could we keep normal FFI context redacted or sanitize URLs and response bodies, with raw details available only through an explicit debug export?

Jasonvdb: Agreed, this folded the request URL (recovery-marker URLs carry a DH-derived private storage path) and the response body into `context`, which is rendered straight into the generated Kotlin/Swift exception.

Fix: `format_context` no longer folds the raw source into `context`. It keeps only the redacted outer label and routes the full cause chain to `tracing::debug!` (developer opt-in, off by default), so raw URLs/paths/bodies stay out of the shipped message. Added `test_sensitive_source_details_never_reach_context`, which plants a private-path URL and a response-body sentinel in the cause chain and asserts neither reaches `context` or `Display`. No enum/field/code change.

paykit-ffi/src/errors.rs
```diff
@@ -69,15 +69,34 @@ impl From<PaykitSdkError> for PaykitFfiError {
...
+// PRIVATE storage path) and a non-2xx HTTP response body. Those are sensitive and
+// must stay out of shipped exception text. Instead we keep only the redacted outer
+// label in `context` and route the full cause chain to `tracing::debug!` — a
+// developer opt-in "debug export" that is not shipped in the user-facing message.
+fn format_context(context: String, source: Option<anyhow::Error>) -> String {
+    if let Some(source) = source {
+        // Debug level is off by default in production builds, so the raw URL / DH
+        // path / response body reaches developers who explicitly enable it but never
+        // leaks into the exception message returned across the FFI boundary.
+        tracing::debug!(
```
ben-kaufman: `tracing::debug!` is not an explicit diagnostics export. It sends the full recovery-marker URL, private path, and response body to any enabled debug logger or subscriber, while Android drops `DEBUG` and the generated Swift/Kotlin API cannot retrieve it. Could we keep raw causes out of logging and expose them only through an explicitly named sensitive diagnostics method?

paykit-ffi/src/errors.rs
```diff
@@ -69,15 +69,34 @@ impl From<PaykitSdkError> for PaykitFfiError {
...
+            cause = %format_args!("{source:#}"),
+            "FFI error cause (redacted from user-facing context)"
+        );
+    }
     context
```
ben-kaufman, line 112: `context` is not always a redacted outer label. `store_encrypted_receipt_json` includes `record.location`, so a failed `issue_receipt` still puts `/pub/paykit/v0/private/.../receipts/{id}` into the generated Swift/Kotlin exception. Could we redact the Receipt Location at that constructor and add a regression test with sensitive data in `context`, not only `source`?

paykit-ffi/src/errors.rs
Jasonvdb: You're right that `tracing::debug!` still ships the raw URL, private path, and body to any subscriber, and Android drops `DEBUG` anyway.

Fix: dropped the raw-cause logging entirely. `format_context` no longer takes or logs the source, so the cause reaches no sink at all and only the redacted outer label survives in `context`. I did not add a sensitive-diagnostics export method, to keep the FFI surface unchanged for this redaction fix. Happy to add an explicitly named one in a follow-up if you want the causes retrievable for debugging. 8c04fcb

paykit-ffi/src/errors.rs
Jasonvdb: Agreed, `store_encrypted_receipt_json` folded `record.location` into `context`, which reaches the generated exception.

Fix: routed that error through a `store_encrypted_receipt_error` constructor that drops the location and keeps a static label, with `test_store_encrypted_receipt_error_redacts_receipt_location` asserting the path is absent from both `context` and `Display`. Also redacted the same leak in the sibling public `paykit-lib` `store_prepared_receipt` (not FFI-reachable today, but the same class in exported code). 8c04fcb

ben-kaufman: Store failures are redacted now, but `retrieve_receipt` still puts `access.location` into `PaykitSdkError::Transport.context` when the Encrypted Receipt is missing. The exported FFI method maps that context directly into the Swift/Kotlin exception, so `/pub/paykit/v0/private/.../receipts/{id}` still crosses the boundary. Could we use a static retrieval context here too and extend the regression test?

Jasonvdb: Confirmed, the missing-receipt branch in `retrieve_receipt` folded `access.location` into `Transport.context`, which the exported FFI method renders verbatim into the generated exception.

Fix: routed it through `missing_encrypted_receipt_error`, a sibling of `store_encrypted_receipt_error` that drops the location and keeps the static label "encrypted receipt was not found at its receipt location", with no source. Extended the regression suite with `test_missing_encrypted_receipt_error_redacts_receipt_location`, asserting the private path is absent from both `context` and `Display`. The persisted record error string keeps the location for local diagnostics: it never crosses the FFI, is redacted from record Debug output, and the record carries the location as a field anyway.

Also audited the other retrieval error paths that feed `last_error`: `fetch_encrypted_receipt_json` ("read encrypted receipt bytes", "fetch encrypted receipt"), `decrypt_receipt`, and `validate_receipt_matches_access` all use static contexts already, and the `RecoveryRequired` fallbacks only echo the caller-supplied `receipt_id`/`counterparty`, never the location. 431b95b

paykit-sdk/src/domain/receipts/mod.rs
```diff
@@ -796,10 +796,43 @@ pub(crate) async fn store_encrypted_receipt_json(
...
+/// (`/pub/paykit/v0/private/.../receipts/{id}`), a PRIVATE storage path whose
+/// folder prefix is DH-derived per counterparty. At the FFI boundary this error's
+/// `context` is rendered verbatim into the generated Kotlin/Swift exception
+/// message, so the Receipt Location MUST NOT be embedded in it. The location is
+/// accepted here only to make the redaction explicit and testable; it is
+/// deliberately dropped, leaving a static, non-sensitive label. The Receipt
+/// Access record the caller persists already carries the location as a field
+/// for local diagnostics.
+pub(crate) fn missing_encrypted_receipt_error(_location: &str) -> PaykitSdkError {
+    PaykitSdkError::Transport {
```
ben-kaufman: This branch only runs after a confirmed 404/GONE, but `missing_encrypted_receipt_error` returns `Transport`, so generated clients receive `transport_error` instead of `not_found`. Could we return `PaykitSdkError::NotFound` with the same static redacted message and update the test to pin that variant?

Jasonvdb: Agreed, the branch only runs after `fetch_encrypted_receipt_json` confirms a 404/GONE, so `Transport` misreported a definitive absence as `transport_error`.

Fix: `missing_encrypted_receipt_error` now returns `PaykitSdkError::NotFound` with the same static redacted label, which crosses the FFI as code `not_found`. `test_missing_encrypted_receipt_error_redacts_receipt_location` pins the variant and the label. `NotFound` is a plain-string variant with no `source`, so no cause chain can reintroduce the path.

The variant change also created a priority question when multiple Receipt Access records fail differently: last-write-wins let a 404 from an older location overwrite a transient transport error from the newest one, surfacing definitive absence where a retry would likely succeed. Added `merge_retrieval_error`: `NotFound` never displaces another error kind and only surfaces when every location 404s, with priority tests.

A self-review pass over the same flow also caught the next sibling: serde_json parse errors embed verbatim document fragments, so a type mismatch while parsing decrypted receipt plaintext (`Receipt::decrypt`) or Receipt Access JSON put decrypted field values into `Protocol`, which crosses the FFI verbatim. Those parse errors are now static labels with no source, `From<PaykitError>` no longer folds `InvalidData` sources into the `Protocol` string, and sentinel regression tests cover both (real AEAD encrypt/decrypt with drifted plaintext). Same fold in the Payment Request and Private Payment List parsers is left for a follow-up, since those paths are being reworked on the PR #104 branch. PR description was stale from before the review iterations and is rewritten to match the shipped behavior. 876eeff

paykit-lib/src/receipt/crypto.rs
```diff
@@ -148,11 +151,12 @@ impl Receipt {
...
-                context: format!("failed to parse receipt plaintext JSON: {err}"),
-                source: Some(err.into()),
-            })?;
+        // SECURITY / REDACTION: the serde error is derived from DECRYPTED
+        // receipt plaintext (its Display embeds field values on type
+        // mismatches), so it must not be folded into the context or kept as
+        // `source` -- this error crosses the FFI boundary as exception text.
+        let receipt_wire: ReceiptWire = serde_json::from_slice(&plaintext)
+            .map_err(|_| invalid_plaintext_json("failed to parse receipt plaintext JSON"))?;
         let receipt = Self::try_from(receipt_wire)?;
```
ben-kaufman, line 160: The new redaction only covers serde failures. A structurally valid decrypted Receipt reaches `Receipt::try_from`, where `validate_wire_version_kind_str` copies the raw `kind` value into `InvalidData.context`, and that context still crosses the FFI verbatim. Could we make the incoming Receipt version/kind error static too and add a sentinel test that passes deserialization but fails this validation?

paykit-ffi/src/errors/tests.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+}
+
+#[test]
+fn test_ffi_sdk_round_trip_recovers_original_via_downcast() {
+    // The four source-bearing arms wrap the original FFI error in the SDK error's
+    // `source`; converting back must downcast to the exact original (same code +
+    // context), not a re-labelled placeholder. Pins F03: without the PaymentAdapter
+    // `source` passthrough the adapter case regresses to `payment_adapter_error`.
+    let originals = [
```
ben-kaufman, line 126: This round-trip test only covers the four source-bearing variants, but the platform callbacks can return any `PaykitFfiError`. `NotFound`, `Protocol`, `Policy`, and `RecoveryRequired` still lose their original context and machine-readable code when converted through the SDK. Could we preserve those variants too, or constrain the callback error surface, and add cases for the remaining four?

paykit-lib/src/receipt/crypto.rs
Jasonvdb: Confirmed, the serde redaction did not cover the validation step after a successful parse: `validate_wire_version_kind_str` echoed the raw `kind` (a decrypted field value) into `InvalidData.context`, which crosses the FFI verbatim.

Fix: the shared validator now uses a static context (`unsupported {label} version/kind`). Every caller of that function validates decrypted private-message plaintext (Receipt, Receipt Access, and all five Payment Request wire parsers), so the fix is at the shared seam rather than special-cased for receipts. The encrypted-envelope version/kind/algorithm check in `Receipt::decrypt` got the same treatment, since those values are fetched-document content. Added `test_decrypt_receipt_wire_validation_error_redacts_plaintext_kind`: a sentinel `kind` that passes deserialization, fails validation through a real AEAD encrypt/decrypt, and is asserted absent from context, source, Display, and Debug. 268884a

paykit-ffi/src/errors/tests.rs
Jasonvdb: Right, `ffi_error_to_sdk` discarded the callback's reason for the four string-variant errors and the custom code survived only inside the outer context.

Fix: `NotFound`, `Protocol`, `Policy`, and `RecoveryRequired` now fold the callback's code and reason into the SDK string (`{context}: {code}: {reason}`). The reason is app-authored and already crossed the boundary once, so echoing it back is not a redaction concern. Added `test_ffi_sdk_round_trip_string_variants_keep_code_and_reason_in_text` covering all four.

What this deliberately does not do: exact downcast recovery. Those SDK variants are plain strings with no `source` field, so the machine-readable custom code degrades to the variant's generic one on the way back (`record_missing` -> `not_found`), which the test pins as the contract. Preserving custom codes losslessly means restructuring the SDK string variants into struct variants with a `source`, a breaking Rust API change that touches every construction site, so I would rather do it as a dedicated follow-up than fold it into this PR. Same for constraining the callback error surface, which is a contract change for existing platform implementations. Happy to open either as a follow-up issue. 268884a

ben-kaufman: Thanks, preserving the reason in the context text is an improvement. The custom code still degrades to the generic variant code, so this does not satisfy the bindings contract to preserve machine-readable error codes across platforms. Could we either preserve the structured code in this PR, or open and link a follow-up issue before resolving this thread?

Jasonvdb: Done in this PR rather than a follow-up issue. The restructure turned out mostly mechanical, so doing the whole thing here was simpler than opening and tracking an issue for it.

Fix: `NotFound`, `Protocol`, `Policy`, and `RecoveryRequired` are now struct variants with the same `{ context, source: Option<anyhow::Error> }` shape as the other four SDK errors. All eight arms of `ffi_error_to_sdk` stash the original callback error in `source`, and the reverse conversion recovers the exact variant, custom code, and reason by downcast. `test_ffi_sdk_round_trip_recovers_original_via_downcast` pins lossless recovery for all eight variants, so no callback code degrades to a generic one.

One tradeoff to note: in the resolved NotFound thread I argued the missing `source` field was itself a defense, since no cause chain could reintroduce a private path. That argument is gone. The guard is now the FFI conversion, which drops every `source` that is not an app-authored `PaykitFfiError` recovered by downcast; paykit-lib and paykit-sdk never construct one, so network-derived or decrypted-plaintext causes can never satisfy the downcast. The new `test_string_variant_sources_never_reach_context` plants a private-path sentinel in each of the four new `source` fields and asserts it reaches neither `context` nor `Display`. Side benefit: `From<PaykitError>` keeps `InvalidData` causes structurally in `Protocol.source` for local diagnostics instead of discarding them, still redacted at the boundary.

This is a breaking Rust API change to `paykit-sdk` (~350 construction/match sites updated in-repo); the UniFFI error shape and generated bindings are unchanged. 907ea04

paykit-sdk/src/error.rs
```diff
@@ -68,13 +92,28 @@ impl From<paykit_lib::PaykitError> for PaykitSdkError {
...
+            // (rendered into generated Kotlin/Swift exception messages), and
+            // lib-level `InvalidData` sources carry raw parse/decode causes
+            // derived from network data or decrypted plaintext. The cause is
+            // kept structurally separate in `source` for local diagnostics.
+            // The FFI conversion drops every `source` except an app-authored
+            // `PaykitFfiError` recovered by downcast; nothing in paykit-lib
+            // or paykit-sdk constructs one, so causes like this one always
+            // miss the downcast and never cross. Only the curated static
+            // `context` label survives in exception text.
+            paykit_lib::PaykitError::InvalidData { context, source } => {
```
ben-kaufman: `PaykitSdkError` derives field-wise `Debug`, so retaining every lib `InvalidData.source` here re-exposes data that this conversion previously dropped. Decrypted Receipt validation can put the offending private field value in that source, which then appears in `format!("{error:?}")` or structured Rust logs. Could we keep dropping lib `InvalidData` sources here, or wrap them in a redacted source type? The callback round-trip still works because those sources are created separately in `paykit-ffi`.

paykit-sdk/src/error.rs
```diff
@@ -35,16 +35,34 @@ pub enum PaykitSdkError {
     },
 
     /// Requested Paykit or Pubky resource was not found.
-    #[error("not found: {0}")]
-    NotFound(String),
+    #[error("not found: {context}")]
+    NotFound {
```
ben-kaufman, line 39: This changes four public `PaykitSdkError` variants from tuple to struct variants, so downstream Rust constructors and matches stop compiling. `CHANGELOG.md` still has an empty `Unreleased` section. Could we add an upgrade note there covering the new variant shape and the callback-code preservation fix?

paykit-sdk/src/error.rs
Jasonvdb: Agreed. Retaining the lib `InvalidData` cause reopened a local logging channel: `PaykitSdkError` derives field-wise `Debug`, so `format!("{err:?}")` and structured Rust logs would render the retained cause even though the FFI conversion never forwards it.

Fix: the conversion drops lib `InvalidData` sources again (`source: None`), with the SECURITY comment now stating the Debug rationale. `test_invalid_data_source_is_not_folded_into_protocol_string` pins the drop and asserts a sentinel cause is absent from both `Display` and `Debug` output. Callback round-trip recovery is unaffected: downcast sources are created only in paykit-ffi's callback plumbing, never from lib errors. ff7c48e

paykit-sdk/src/error.rs
Jasonvdb: Added the Unreleased upgrade note. Changed covers the breaking tuple-to-struct shape change for the four variants with the new construction and pattern forms (generated Swift/Kotlin error shape unchanged), the `not_found` contract change for missing Encrypted Receipts, and the redacted exception text scoped to what this PR actually fixes (cause chains, Receipt Locations, Receipt/Receipt Access/version-kind labels). Fixed covers the lossless callback error round trip. ff7c48e

ben-kaufman: Thanks, the upgrade note is there now, but `so decrypted plaintext never reaches generated Swift/Kotlin exception text` is broader than the current code. The PR description still lists Payment Request and Private Payment List serde errors as a known plaintext follow-up. Could we narrow this to say Receipt, Receipt Access, and version/kind values no longer reach exception text? `InvalidData` sources are also dropped during lib-to-SDK conversion now, so it would be useful to mention that Rust SDK errors lose those causes before the FFI boundary.

Jasonvdb: Narrowed as suggested: the bullet now says Receipt, Receipt Access, and version/kind values no longer reach exception text, instead of the blanket decrypted-plaintext claim; Payment Request and Private Payment List parse detail stays listed as the known follow-up in the PR description.

Also added the Changed entry for Rust consumers: errors converted from `paykit_lib::PaykitError::InvalidData` drop the underlying cause during lib-to-SDK conversion, so `Error::source()` returns `None` and `Debug` output no longer includes the parse/decode cause. e62484d

ben-kaufman: Thanks, this is closer, but two claims are still broader than current behavior. Private Payment List parsing and outbound private-message validation still put raw version/kind values into `Protocol.context`, so those values can still reach FFI exception text. Also, dropping `InvalidData.source` removes the separate source chain, but some lib errors copy the same parse/decode detail into `context`, which SDK `Debug` still prints. Could we narrow the first bullet to the specific Receipt, Receipt Access, and shared validation paths covered here, and say the second change drops the structured source chain rather than all parse/decode cause detail?

Jasonvdb: Both points were right, and checking them against the code turned up real gaps on the Receipt path itself, so this round is code plus wording rather than wording alone.

Fix: the remaining dynamic contexts on the Receipt path are now static: the encrypted-envelope base64 and nonce-length decode errors (`receipt/crypto.rs`; base64 `DecodeError` Display names the offending byte and offset of the fetched document), Receipt Decryption Key validation (`receipt/types.rs`; detail dropped entirely rather than kept in `source`, since it can name a byte of the key text), the fetched-body UTF-8 check (`domain/receipts/mod.rs`), and the backup-restore `err.to_string()` fold over `decrypt_receipt` (`backup/validation.rs`). The decode tests now pin exact equality instead of `contains`, so a reintroduced `: {err}` suffix fails.

The CHANGELOG redaction bullet now enumerates the exact redacted paths instead of the categorical claim, and names Private Payment List parsing and the SDK's outbound private-message validation (`validate_outbound_private_message`) as remaining echoes for the follow-up PR. The `InvalidData` bullet now says the conversion drops the structured source chain, and that detail folded into `context` by lib call sites still appears in `Display`/`Debug`. The PR description's scope claims were re-checked against the code the same way and adjusted to match. c1bee01

Jasonvdb:
This PR escalated beyond my original quick fix but I think it's good now

ben-kaufman (approved):

PR #99, opened 2026-07-14
Zeroize ReceiptDecryptionKey on drop

Jasonvdb:
## Problem

`ReceiptDecryptionKey` (a 256-bit AEAD key, `paykit-lib/src/receipt/types.rs`) was
the workspace outlier: it held key material with **no zeroization**, while
paykit-sdk and paykit-ffi zeroize their secrets. Transient decoded copies also
lingered in `generate()`/`new()`/`bytes()` and in the `[u8; 32]` used to build the
cipher.

## Changes

- **`impl Drop`** scrubs the base64url `String`. The existing
  `#[derive(Clone, PartialEq, Eq)]` are kept (`Eq` is required by upstream
  derives); clones zeroize independently, which is correct.
- **Transient buffers scrubbed** in `generate()` (raw key `GenericArray` slice),
  `new()` (decoded length-check buffer), and `bytes()` (decoded buffer — the
  returned `[u8; 32]` is copied out first, then the buffer is scrubbed).
- **`receipt/crypto.rs`** scrubs the `[u8; 32]` key array immediately after cipher
  construction in both `encrypt_for_location` and `decrypt` (the cipher holds its
  own key schedule).
- **`zeroize = "1"`** added to paykit-lib — already in the tree via
  `chacha20poly1305`, so **no new external code** (`Cargo.lock` gains only the dep
  edge).

## Honest limitation (auditor note in code)

Zeroization here is **defense-in-depth for in-memory copies, not a complete
scrub**: the key is serialized verbatim into the Receipt Access wire JSON, so
copies necessarily exist outside this type (network buffers, serialized messages,
homeserver storage).

## Tests

- `new()` rejects non-32-byte and non-base64url material (scrub doesn't disturb
  validation).
- `bytes()` round-trips (scrub doesn't corrupt the key).
- A clone survives the original being dropped/zeroized (independent per-clone
  scrubbing).

## Protocol / API impact

- **No public API change**; **no wire/serde change** — serialization still goes
  through `as_str()`. Protocol impact: none.

## Verification (local)

- `cargo test -p paykit-lib` (receipt round-trip + new key tests pass)

### DIFF
--- a/paykit-lib/src/receipt/types.rs
+++ b/paykit-lib/src/receipt/types.rs
@@ -3,7 +3,7 @@ use std::fmt;
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine as _};
 use chacha20poly1305::{aead::OsRng, KeyInit, XChaCha20Poly1305};
 use serde_json::{Map as JsonMap, Value as JsonValue};
-use zeroize::Zeroize;
+use zeroize::{Zeroize, Zeroizing};
 
 use crate::{
     validation::validate_uuid_v4, BillingPeriod, EventId, PaykitError, PaymentAmount,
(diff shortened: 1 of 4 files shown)


paykit-lib/src/receipt/types.rs
```diff
@@ -134,26 +135,36 @@ impl fmt::Debug for Receipt {
...
+        // is the only surviving copy, guarded by this type's Drop impl.
+        key.as_mut_slice().zeroize();
+        Self(encoded)
     }
 
     /// Validate and construct a Receipt Decryption Key from base64url text.
     pub fn new(key: impl Into<String>) -> Result<Self> {
         let key = key.into();
-        let bytes = URL_SAFE_NO_PAD.decode(&key).map_err(|err| {
+        let mut bytes = URL_SAFE_NO_PAD.decode(&key).map_err(|err| {
```
ben-kaufman: `new()` still drops candidate key material normally on its error paths. `base64::decode` can write a decoded prefix before returning an error, and the owned input `String` also bypasses the new `Drop` because construction never succeeds. Could we keep both the input and a caller-owned decode buffer under zeroizing guards across every return path?

Jasonvdb: Good point, the error paths left candidate key material unscrubbed.

Fix: `new()` now decodes into a `Zeroizing<Vec<u8>>` (so any prefix a failing decode wrote is scrubbed on drop) and zeroizes the input `String` before every early return. Success still moves the key into the Drop-guarded type; messages and behavior are unchanged.

ben-kaufman (approved):

PR #100, opened 2026-07-14
Redact persisted error strings in record Debug output

Jasonvdb:
## Problem

In `paykit-sdk/src/storage/records.rs`, four persisted error strings were printed
**raw** in `Debug` output while sibling record types redact theirs. The two
private-stream `parse_error` fields derive from private plaintext, making them the
most sensitive.

## Changes (all within `records.rs` + its test)

- `PublicEndpointRecord.last_error` — redacted (its manual `Debug` already redacted
  `payload`, but not the error).
- `LinkedPeerRecord.local_recovery_marker_last_error` — replaced the **derived**
  `Debug` with a manual impl that mirrors every field and redacts only the error.
- `NewPrivateStreamItem.parse_error` and `PrivateStreamItemRecord.parse_error` —
  redacted.
- Each redaction mirrors the existing `<redacted:N bytes>` style exactly (as used
  by `OutboundPrivateMessageRecord.last_error`).

## Tests

- Extended `test_sensitive_storage_debug_is_redacted` with fixtures that populate
  each of the four fields with a unique `*-secret` marker and assert it does not
  appear in the `Debug` output — fixtures that would have caught this. The test now
  exercises both `NewPrivateStreamItem` and `PrivateStreamItemRecord` (the shared
  `parse_error` field) plus `LinkedPeerRecord` and `PublicEndpointRecord`.

## Scope / impact

- **Not a live leak:** backups are documented as caller-encrypted, so this is a
  Debug-consistency / log-hygiene fix.
- Kept the diff to one `impl` block per type (the `LinkedPeerRecord` manual impl is
  a minor conflict surface with in-flight recovery-marker work — trivial to rebase).
  **Protocol impact: none.**

## Verification (local)

- `cargo test -p paykit-sdk --lib` (redaction test passes)

### DIFF
--- a/paykit-sdk/src/storage/records.rs
+++ b/paykit-sdk/src/storage/records.rs
@@ -15,7 +15,7 @@ use crate::{
 };
 
 /// Durable Linked Peer state.
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct LinkedPeerRecord {
     /// Counterparty public key.
     pub counterparty: PubkyPublicKey,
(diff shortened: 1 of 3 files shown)


paykit-sdk/src/storage/records.rs
```diff
@@ -42,6 +42,42 @@ pub struct LinkedPeerRecord {
...
+            .field("last_sync_at", &self.last_sync_at)
+            .field("last_private_receive_at", &self.last_private_receive_at)
+            .field("failure_count", &self.failure_count)
+            .field("local_recovery_attempt_id", &self.local_recovery_attempt_id)
+            .field(
+                "local_recovery_marker_created_at",
+                &self.local_recovery_marker_created_at,
+            )
+            .field(
+                "local_recovery_marker_last_error",
```
ben-kaufman, line 63: I think the same error still leaks through `EncryptedLinkRecoveryMarkerReport`. `from_peer` copies this value into `local_marker_last_error`, and that public report still derives `Debug`, so `format!("{report:?}")` prints the raw string. Could we give the SDK report a redacting `Debug` implementation and add a regression assertion for it too?

Jasonvdb: Confirmed, `EncryptedLinkRecoveryMarkerReport` derived `Debug` and `from_peer` copied the raw error in, so `{report:?}` printed it.

Fix: replaced the derive with a manual `Debug` that renders `local_marker_last_error` as `<redacted:N bytes>` (matching the `LinkedPeerRecord` idiom), leaving the non-sensitive fields intact. Added `test_recovery_marker_report_debug_redacts_local_marker_error`, which plants a sentinel error via `from_peer` and asserts it is absent from the debug output. 7440e7c

ben-kaufman (approved):
Re-audited at 7440e7c. The reported Debug redaction issue is fixed, and the CI checks are green.

PR #101, opened 2026-07-14
Handle unavailable Android certificate revocation status

ben-kaufman:
## Summary

- vendor the Android support artifact used by `rustls-platform-verifier` 0.6.2
- distinguish explicit certificate revocation from unavailable revocation status
- keep unrelated certificate-validation failures fail-closed
- rebuild all Android native bindings with the corrected verifier artifact
- prepare the Rust, Swift, and Android packages as `0.1.0-rc36`

## Motivation

On Android 16 / API 36, Pubky Auth relay delivery to `httprelay.pubky.app` can fail because Android/Bouncy Castle reports missing CRL data as a recoverable `CertPathValidatorException`. The upstream Android support artifact maps every exception of that type to `Revoked`, even when revocation was not established.

This blocks Pubky Auth before normal approval despite the relay certificate not being reported as explicitly revoked.

## Security behavior

- `BasicReason.REVOKED` remains a hard failure.
- `UNDETERMINED_REVOCATION_STATUS` and the known Android `No CRLs found for issuer` condition follow the verifier's existing best-effort revocation policy.
- Other unexpected validator errors return `UnknownCert` and fail closed.
- Chain, hostname, validity, signature, and EKU validation are unchanged.

The vendored directory records the exact upstream version, source patch, tests, licenses, build command, and AAR checksum.

## Verification

- patched verifier Kotlin unit tests: passed
- patched verifier release AAR build: passed
- Android 16 / API 36 instrumented auth-relay TLS smoke test: passed against the default relay
- final Paykit Android AAR inspected to confirm it packages the patched verifier classes
- `cargo fmt --check`: passed
- `cargo clippy --all-targets --all-features`: passed
- `cargo doc --no-deps`: passed
- Paykit SDK unit tests: 366 passed
- `cd paykit-ffi && ./build.sh --release --rc all`: passed for iOS and all Android ABIs

The testnet suites remain blocked on this machine before test logic by embedded PostgreSQL shared-memory exhaustion. Two host-only FFI Pubky Auth tests also fail identically on `origin/master` because the Pubky client cannot initialize in this macOS environment.

This PR prepares the `0.1.0-rc36` package metadata and artifacts. It does not
create a tag, publish a GitHub release, or publish either platform package.


vendor/rustls-platform-verifier-android/android-revocation-soft-fail.patch
```diff
@@ -0,0 +1,91 @@
...
++    error.reason == CertPathValidatorException.BasicReason.REVOKED
++
++internal fun isRevocationStatusUnavailable(error: CertPathValidatorException): Boolean {
++    if (error.reason == CertPathValidatorException.BasicReason.UNDETERMINED_REVOCATION_STATUS) {
++        return true
++    }
++
++    return error.reason == CertPathValidatorException.BasicReason.UNSPECIFIED &&
++        generateSequence<Throwable>(error) { it.cause }
++            .any { it.message?.startsWith("No CRLs found for issuer ") == true }
```
Jasonvdb, line 16: The `"No CRLs found for issuer "` prefix match is load-bearing but its provenance isn't recorded anywhere. Android reports the missing-CRL case as `BasicReason.UNSPECIFIED`, so the `UNDETERMINED_REVOCATION_STATUS` branch doesn't cover it, and the string is an internal Bouncy Castle message that Android can reword in any release. If that happens, verification quietly falls back to `UnknownCert` (fail closed) and the relay failure this PR fixes comes back with nothing pointing at the cause. Fix: add a short note here (or in vendor/rustls-platform-verifier-android/README.md) recording that the string is the Bouncy Castle wording observed on current Android releases (rustls-platform-verifier#221) and that a future reword will surface as `StatusCode.UnknownCert` with the BC message in the error string. No behavior change needed.

Jasonvdb (approved):

vendor/rustls-platform-verifier-android/android-revocation-soft-fail.patch
ben-kaufman: Added this context to the vendored verifier README, including the Bouncy Castle provenance and the fail-closed behavior if Android changes the message. No verifier behavior changed.

PR #103, opened 2026-07-15
test(receipt): cover decrypt_receipt AEAD failure paths

Jasonvdb:
## Problem

`decrypt_receipt` (`paykit-lib/src/receipt/crypto.rs`) had **zero**
malformed-envelope coverage. The two security-critical branches — wrong key
and tampered ciphertext — were untested, so a refactor that let a wrong key
or a tampered ciphertext decrypt successfully would pass the whole suite. The
envelope-validation branches (nonce length, base64url decoding, envelope
header, envelope JSON parse) were likewise unpinned.

## Change (tests only)

Add 7 failure-path tests plus one small helper to the existing
`#[cfg(test)] mod tests` in `paykit-lib/src/receipt.rs`, reusing
`prepared_receipt_for_test` and the module's
`matches!(err, PaykitError::InvalidData { .. })` assertion style:

| Test | Pins |
|------|------|
| `test_decrypt_receipt_wrong_key_rejected` | **Key binding** — key A ciphertext must not decrypt under key B |
| `test_decrypt_receipt_tampered_ciphertext_rejected` | **AEAD integrity** — one flipped byte fails the Poly1305 tag |
| `test_decrypt_receipt_short_nonce_rejected` | nonce-length guard (`nonce must be 24 bytes`) |
| `test_decrypt_receipt_non_base64url_ciphertext_rejected` | ciphertext base64url decode |
| `test_decrypt_receipt_non_base64url_nonce_rejected` | nonce base64url decode |
| `test_decrypt_receipt_unsupported_envelope_version_rejected` | envelope header/version check |
| `test_decrypt_receipt_malformed_envelope_json_rejected` | envelope JSON parse |

Each test asserts `InvalidData` **and** the specific error-context substring
emitted by the branch under test, so it documents *which* check fired and
cannot pass by tripping a different (earlier) branch. The two crown-jewel
tests (wrong key, tampered ciphertext) both use the *correct* key/envelope
otherwise, isolating the AEAD failure they target.

The `tamper_encrypted_envelope` helper mutates a single existing envelope
field and re-serializes. Because it only touches fields that already exist,
the wire type's `deny_unknown_fields` still accepts the JSON, so every
tampered payload reaches the exact decrypt branch it asserts.

## Scope / impact

- **Tests only**; no production code, no public API, no wire/serde change.
  **Protocol impact: none.**
- All in the existing test module (`receipt.rs`); no hot zones touched.

## Verification (local)

- `cargo test -p paykit-lib --lib receipt::tests` — **24 passed** (17 baseline + 7 new)

### DIFF
--- a/paykit-lib/src/receipt.rs
+++ b/paykit-lib/src/receipt.rs
@@ -380,6 +380,31 @@ mod tests {
     }
 
     #[test]
+    fn test_decrypt_receipt_overlong_nonce_rejected() {
+        // Pins the other side of the `nonce.len() != 24` guard: a 25-byte nonce
+        // must be rejected before the fixed-size XNonce conversion, which would
+        // otherwise panic. A `< 24` guard would let this case through.
+        let prepared = prepared_receipt_for_test();
+        let tampered = tamper_encrypted_envelope(&prepared.encrypted_receipt, |obj| {
+            let mut nonce = URL_SAFE_NO_PAD
+                .decode(obj.get("nonce").and_then(JsonValue::as_str).unwrap())
+                .unwrap();
+            nonce.push(0);
+            obj.insert(
+                "nonce".to_string(),
+                JsonValue::String(URL_SAFE_NO_PAD.encode(&nonce)),
+            );
+        });
+
+        let err = decrypt_receipt(&tampered, &prepared.access.key, &prepared.access.location)
+            .unwrap_err();
+        assert!(
+            matches!(err, PaykitError::InvalidData { ref context, .. } if context.contains("nonce must be 24 bytes")),
+            "expected overlong-nonce rejection, got: {err}"
+        );
+    }
+
+    #[test]
     fn test_decrypt_receipt_non_base64url_ciphertext_rejected() {
         let prepared = prepared_receipt_for_test();
         let tampered = tamper_encrypted_envelope(&prepared.encrypted_receipt, |obj| {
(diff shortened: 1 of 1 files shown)


paykit-lib/src/receipt.rs
```diff
@@ -297,6 +297,155 @@ mod tests {
...
+            matches!(err, PaykitError::InvalidData { ref context, .. } if context.contains("nonce is not valid base64url")),
+            "expected non-base64url nonce rejection, got: {err}"
+        );
+    }
+
+    #[test]
+    fn test_decrypt_receipt_unsupported_envelope_version_rejected() {
+        let prepared = prepared_receipt_for_test();
+        let tampered = tamper_encrypted_envelope(&prepared.encrypted_receipt, |obj| {
+            obj.insert("version".to_string(), JsonValue::from(2u8));
```
ben-kaufman, line 447: Could we add cases that replace `kind` and `algorithm` too? This only pins the `version` part of the combined header guard, so removing either other comparison leaves the receipt tests green and lets a mislabeled envelope decrypt.

paykit-lib/src/receipt.rs
```diff
@@ -297,6 +297,155 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_decrypt_receipt_short_nonce_rejected() {
+        let prepared = prepared_receipt_for_test();
+        let tampered = tamper_encrypted_envelope(&prepared.encrypted_receipt, |obj| {
+            let mut nonce = URL_SAFE_NO_PAD
+                .decode(obj.get("nonce").and_then(JsonValue::as_str).unwrap())
+                .unwrap();
+            nonce.truncate(12);
```
ben-kaufman, line 367: Could we add an overlong nonce case too? The 12-byte case only pins one side of `nonce.len() != 24`. If that guard changes to `< 24`, the receipt tests stay green and a 25-byte nonce reaches the fixed-size conversion and panics.

paykit-lib/src/receipt.rs
Jasonvdb: Right, only the version arm of the combined guard was pinned.

Fix: added `test_decrypt_receipt_wrong_kind_rejected` and `test_decrypt_receipt_wrong_algorithm_rejected`, each asserting the "unsupported encrypted receipt envelope" rejection. Deleting either the `kind` or `algorithm` comparison in `crypto.rs` now turns a test red. 6faf29a

paykit-lib/src/receipt.rs
Jasonvdb: Good catch, the 12-byte case only pinned the short side.

Fix: added `test_decrypt_receipt_overlong_nonce_rejected` with a 25-byte nonce, asserting the "nonce must be 24 bytes" rejection. If the guard changed from `!= 24` to `< 24` this test fails, so an overlong nonce can never reach the fixed-size `XNonce` conversion. 6faf29a

ben-kaufman (approved):

PR #104, opened 2026-07-15
test(payment-request): add proptest coverage for the wire format

Jasonvdb:
## Problem

There was **no property-based testing anywhere in the workspace**. The
payment-request wire format (`paykit-lib/src/payment_request/wire.rs`) is the
highest-value first target: money-adjacent, network-facing, hand-rolled
parse/serialize logic where example-based tests only cover the shapes someone
thought of.

## Change

- Add `proptest = "1"` to `paykit-lib` `[dev-dependencies]` (test-only). The
  `Cargo.lock` diff is limited to proptest and its transitive deps — no
  removals, no version bumps to existing packages.
- New `paykit-lib/src/tests/payment_request_properties.rs` (registered in
  `tests/mod.rs`) with 4 properties, 256 cases each (override via
  `PROPTEST_CASES`):
  - **`valid_event_round_trips`** — generate construction-valid
    `PaymentRequestEvent` values, `serialize_payment_request_event` →
    `parse_payment_request_event_message`, and assert the parsed event equals
    the original.
  - **`arbitrary_strings_never_panic`** / **`arbitrary_json_never_panics`** —
    feed fuzzed strings and bounded arbitrary JSON through the public parser;
    assert it returns `Some`/`None` (and every accessor + `Debug` path runs)
    without panicking.
  - **`kinded_json_never_panics`** — JSON carrying a recognized payment-request
    `kind` routes through `known_kind` into the per-kind parsers, exercising
    the deepest parse paths.
- `test_payment_request_routing_covers_all_private_message_kinds` drives every
  `PrivateMessageKind` variant through the parser. Cases, expected routing, and
  the proptest kind strings all expand from a single macro-generated
  declaration with a wildcard-free exhaustiveness match, so a new enum variant
  fails to compile until classified and then flows into the tests
  automatically.

Every value strategy emits only construction-valid data (valid UUID-v4 ids,
`Z`-suffixed RFC3339 timestamps, decimal amounts, non-reserved endpoint
identifiers, `ends_at`-after-`starts_at` billing periods) and documents where
it is deliberately **narrower than the full domain** so the round-trip never
fails on a bad generator rather than a real bug.

## Review follow-up: exhaustive kind routing (production change)

Per review, the original tests-only guard still allowed a future routed kind to
slip past two hand-maintained lists while `parse_event` panicked on
`unreachable!()`. Commit aff3291 fixes the root cause:

- `parse_event` now returns `Option<Result<PaymentRequestEvent>>` and matches
  every `PrivateMessageKind` with no wildcard arm. Unrouted kinds return
  `None` and the `unreachable!()` panic path is removed. Adding an enum
  variant fails to compile until it is explicitly routed or rejected.
- `is_payment_request_event` (only caller was the dispatcher guard) is folded
  into that single match, so routing has one source of truth.
- No public API changes: `parse_event` is private and the removed predicate
  was `pub(crate)`, so downstream bindings are unaffected.

## Adding strategies for future wire types

To extend this to another wire type: write a strategy that emits only
construction-valid *typed* values (build the domain type, not raw JSON), then
assert `parse(serialize(value)) == value`; pair it with a never-panic property
pushing arbitrary bytes/JSON through the public parser. Keep valid-value
strategies narrow at first and widen them as the parser's guarantees are
pinned. The module header documents this pattern inline.

## Verification (local)

- `cargo test -p paykit-lib --lib` — **185 passed**
- Compile-break check: adding a hypothetical `FutureKind` variant errors in
  `parse_event`; adding it plus a production route without updating the test
  declaration errors in the macro's exhaustiveness guard.


paykit-lib/src/tests/payment_request_properties.rs
```diff
@@ -0,0 +1,332 @@
...
+        ]
+    })
+}
+
+/// Build a JSON object carrying a recognized Payment Request `kind` so the
+/// parser routes past `known_kind` into the per-kind parsers (exercising the
+/// deepest code paths, including the `unreachable!()` invariant).
+fn arb_kinded_json() -> impl Strategy<Value = String> {
+    (
+        prop::sample::select(PAYMENT_REQUEST_EVENT_KINDS),
```
ben-kaufman: This only samples the five kinds that already have `parse_event` match arms, so it cannot catch `is_payment_request_event()` accepting another kind while the dispatcher still falls through to `unreachable!()`. Could we drive this from all `PrivateMessageKind` variants and assert that every kind selected by `is_payment_request_event()` has a non-panicking route?

Jasonvdb: Agreed, the hardcoded five-string list was decoupled from the enum, so a new kind that `is_payment_request_event()` accepts without a `parse_event` arm would slip past.

Fix: added `test_payment_request_routing_covers_all_private_message_kinds`, driven from an enum-typed list of every `PrivateMessageKind` and guarded by a wildcard-free match (adding a variant fails to compile until it is classified). It pushes each kind through `parse_payment_request_event_message` asserting no panic, and cross-checks the routed set against `is_payment_request_event()` and `PAYMENT_REQUEST_EVENT_KINDS`. 8062b6f

ben-kaufman: This catches selector drift for the current enum, but the future-variant case is still decoupled. `payment_request_routing_expectation` is exhaustive, while `ALL_PRIVATE_MESSAGE_KINDS` and `PAYMENT_REQUEST_EVENT_KINDS` are separate manual lists, and the final assertion checks only the count. I added a new routed variant, updated the enum mappings, predicate, and expectation, but left it out of both lists. All five tests passed, while parsing that kind still hit `unreachable!()`. Could we generate the cases, routing expectation, and kind strings from one exhaustive declaration, or make `parse_event` exhaustive?

Jasonvdb: Confirmed, your scenario reproduced here: with the expectation match updated but both lists untouched, all five tests passed while the new kind panicked in `parse_event`. The count-only cross-check was the weak link.

Fix: took both halves of your suggestion in aff3291.

- `parse_event` (api.rs:27) now returns `Option<Result<PaymentRequestEvent>>` and matches every `PrivateMessageKind` with no wildcard arm. Unrouted kinds return `None`, the `unreachable!()` is gone, and a new variant fails to compile until it is routed or rejected there. `is_payment_request_event` had no other caller, so it is folded into that single match instead of being a second predicate that could drift.
- The tests expand one declaration (`declare_private_message_kind_routing!`, payment_request_properties.rs:53) into the `(kind, routed)` cases, a wildcard-free exhaustiveness match, and, via `routed_kind_strings`, the proptest kind strings. The two manual lists and the count check are deleted; there is no second place to forget.

Verified both failure modes are now compile errors: adding a `FutureKind` variant alone breaks at `parse_event`, and adding it plus a production route without touching the test declaration breaks at the macro's exhaustiveness guard. Once both compile, the new kind flows into the routing test and `kinded_json_never_panics` automatically.

paykit-lib/src/tests/payment_request_properties.rs
```diff
@@ -0,0 +1,412 @@
...
+        // The parser reads the kind from the raw JSON body via `known_kind`, so
+        // carry it there rather than in the struct's `kind` header field.
+        let message = PrivateApplicationMessage {
+            version: None,
+            kind: None,
+            raw_json: format!(r#"{{"kind":"{}"}}"#, kind.as_str()),
+        };
+
+        // Touch every accessor so a panic anywhere downstream fails this test.
+        let parsed = parse_payment_request_event_message(&message);
```
ben-kaufman, line 405: Could we assert `message.known_kind() == Some(kind)` before checking `parsed.is_some()`? For an ignored kind, a missing or wrong `PrivateMessageKind::parse` mapping also returns `None`, so this passes without exercising that variant. I reproduced it by adding an ignored variant everywhere except `PrivateMessageKind::parse`; all five property tests still passed.

Jasonvdb: Confirmed. For an ignored kind the routing assertion could not distinguish "recognized and ignored" from "never mapped in `PrivateMessageKind::parse`", both return `None` from the public parser, so your repro passed without exercising the variant.

Fix: the routing test now asserts `message.known_kind() == Some(kind)` before checking routing, pinning the `as_str` -> `parse` round-trip for every variant. Replayed your repro (added an ignored variant everywhere except `parse`): the test now fails with "PrivateMessageKind::parse does not round-trip paykit.future_kind". ef37147

Also resolved the merge conflict with master in 5c9d5c5: PR #97 fixed the same `parse_event` panic by returning `InvalidData` from explicit non-request arms behind the `is_payment_request_event` guard. Kept the single exhaustive `Option` match from this thread instead (public behavior identical, no second predicate to drift, and no error string on that path that could carry decrypted plaintext) and adapted master's leak test to it as `test_parse_event_non_request_kind_ignored`.

ben-kaufman (approved):

PR #105, opened 2026-07-15
test(validation): cover UUID-v4 version/variant rejection

Jasonvdb:
## Problem

`validate_uuid_v4` (`paykit-lib/src/validation.rs`) underpins `EventId`,
`PaymentRequestId`, `ReceiptId`, and recovery markers, but its **version/variant
rejection branch** (`get_version_num() != 4 || get_variant() != RFC4122`) had
**zero** coverage. The only existing rejection test feeds a non-UUID string,
which fails earlier at `Uuid::try_parse` — the parse branch, not the
version/variant branch. Nil, v1, v3, and v5 UUIDs (well-formed but not v4) were
untested everywhere.

## Change (tests only)

New `paykit-lib/src/tests/event_id.rs` (4 tests) exercising the validator
through the public id newtypes. Scoped to behavior **this crate owns** — the
`uuid` crate's own version detection is not re-tested:

- **Canonicalization:** an uppercase v4 UUID normalizes to the stored lowercase
  form (also covers acceptance of a valid v4). This matters because
  `PartialEq`/`Hash` derive on the raw `String`, so two case-variant spellings
  of the same UUID must not compare/hash as distinct ids.
- **Rejection:** one parametrized test over parseable-but-non-v4 UUIDs (nil,
  v1, v3, v5) asserting each returns `PaykitError::Validation(_)` — the
  previously-uncovered version/variant branch.
- **Through the newtypes:** accept + reject pairs via `PaymentRequestId::new`
  and `ReceiptId::new`, pinning that those newtypes route through the same
  validator rather than storing arbitrary strings.

## Placement note

The tests live under `src/tests/` rather than inline. The `tests/README`
convention keeps small pure unit tests next to their module but routes tests
that **"drive multiple Paykit modules together"** into `src/tests/`. This file
exercises one shared validator through three id newtypes living in three
different modules (`event.rs`, `payment_request/types.rs`, `receipt/types.rs`),
so there is no single inline home; a dedicated cross-cutting file is the natural
fit. Easy to relocate if a reviewer prefers otherwise.

## Scope / impact

- **Tests only**; no production code, no public API change. **Protocol impact:
  none.**
- Pure synchronous tests — no testnet, no shared fixtures.

## Verification (local)

- `cargo test -p paykit-lib --lib event_id` — **4 passed**

### DIFF
--- a/paykit-lib/src/tests/event_id.rs
+++ b/paykit-lib/src/tests/event_id.rs
@@ -1,0 +1,87 @@
+//! Unit tests for the shared UUID-v4 validator (`validate_uuid_v4`), exercised
+//! through the public `EventId`, `PaymentRequestId`, and `ReceiptId`
+//! constructors.
+//!
+//! These pin the version/variant rejection branch (nil, v1, v3, v5) that no
+//! other test covers, plus the acceptance and lowercase-canonicalization paths.
+
+use super::*;
+
+/// A valid, lowercase, RFC4122 version-4 UUID.
+const VALID_V4: &str = "550e8400-e29b-41d4-a716-446655440000";
+
+/// The same UUID as `VALID_V4`, spelled in uppercase.
+const VALID_V4_UPPERCASE: &str = "550E8400-E29B-41D4-A716-446655440000";
+
+/// Nil UUID: version 0, non-RFC4122 variant.
+const NIL_UUID: &str = "00000000-0000-0000-0000-000000000000";
+
+/// Time-based (version 1) UUID.
+const V1_UUID: &str = "a8098c1a-f86e-11da-bd1a-00112444be1e";
+
+/// Name-based MD5 (version 3) UUID.
+const V3_UUID: &str = "6fa459ea-ee8a-3ca4-894e-db77e160355e";
+
+/// Name-based SHA-1 (version 5) UUID.
+const V5_UUID: &str = "886313e1-3b8a-5372-9b90-0c9aee199e5d";
+
+#[test]
+fn test_event_id_accepts_valid_v4() {
+    let id = EventId::new(VALID_V4).expect("a valid v4 UUID must be accepted");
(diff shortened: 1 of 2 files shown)


paykit-lib/src/tests/event_id.rs
```diff
@@ -0,0 +1,69 @@
...
+/// A valid, lowercase, RFC4122 version-4 UUID.
+const VALID_V4: &str = "550e8400-e29b-41d4-a716-446655440000";
+
+/// The same UUID as `VALID_V4`, spelled in uppercase.
+const VALID_V4_UPPERCASE: &str = "550E8400-E29B-41D4-A716-446655440000";
+
+/// Well-formed UUIDs that parse successfully but are not RFC4122 version 4, so
+/// they must fail the version/variant check rather than the parse step: the nil
+/// UUID (version 0) plus versions 1, 3, and 5.
+const NON_V4_UUIDS: &[&str] = &[
```
ben-kaufman, line 32: Could we add a version-4 UUID with a non-RFC4122 variant here? Every current fixture fails `get_version_num() != 4` first, so the variant check is still untested. For example, `550e8400-e29b-41d4-0716-446655440000` parses as version 4 with the NCS variant.

Jasonvdb: Right, every fixture failed the version check first so the variant branch was untested.

Fix: added `550e8400-e29b-41d4-0716-446655440000` (version-4 nibble, NCS variant) to `NON_V4_UUIDS`, plus `test_event_id_rejects_v4_with_non_rfc4122_variant`, which first asserts the fixture parses as v4 with a non-RFC4122 variant and then asserts `EventId::new` rejects it with `Validation`. Dropping the variant clause from `validate_uuid_v4` now turns it red. 91e8e51

ben-kaufman (approved):

PR #108, opened 2026-07-16
test(encrypted-link): pin the 1000-byte noise message boundary

Jasonvdb:
## Problem

The send-time size ceiling in
`paykit-lib/src/encrypted_link/private_application_message.rs`
(`plaintext.len() > PUBKY_NOISE_MSG_LEN`) was only tested at `LEN+1`. There
was no **at-limit acceptance** test, so a `>` → `>=` off-by-one regression
(rejecting exactly-1000-byte messages) would pass the suite. There was also no
realistic test of a Private Payment List crossing the ceiling via many small
Payment Endpoints.

## Change (tests only)

Two tests inside the file's existing `#[cfg(test)]` module (the validator is
module-private, so they cannot live in the centralized `tests/` directory):

- **At-limit acceptance:** a payload of exactly `PUBKY_NOISE_MSG_LEN` bytes is
  accepted (uses the constant, not a literal `1000`).
- **Many-small-endpoints rejection:** a real `PrivatePaymentList` of 20
  endpoints built via the validated constructors serializes past the ceiling
  (sanity-asserted in the test) and is rejected with
  `PaykitError::Validation`, using the real send-path context string.

No construction-time size enforcement was added — send-time validation is the
documented design.

## Scope / impact

- **Tests only**; no production code. **Protocol impact: none.**
- Pure synchronous unit tests.

## Verification (local)

- `cargo clippy -p paykit-lib --all-targets --all-features -- -D warnings`
- `cargo test -p paykit-lib --lib private_application_message` — **16 passed**

### DIFF
--- a/paykit-lib/src/encrypted_link/private_application_message.rs
+++ b/paykit-lib/src/encrypted_link/private_application_message.rs
@@ -349,6 +349,11 @@ pub(super) async fn send_private_application_message(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::{
+        serialize_private_payment_list_json, PaymentEndpointIdentifier, PaymentEndpointPayload,
+        PrivatePaymentList,
+    };
+    use std::collections::HashMap;
 
     #[test]
     fn test_send_attempts_from_retries_bounds() {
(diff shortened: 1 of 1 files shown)


ben-kaufman (approved):
Reviewed at the current head. No actionable issues found.

PR #109, opened 2026-07-16
test(payment-amount): cover validation failure paths

Jasonvdb:
## Problem

`PaymentAmount` validation had untested failure branches: the
control-character asset check (`char::is_control`) and malformed decimal
shapes. The currently-accepted bare-decimal-point forms (`".5"`, `"10."`) were
also unpinned, so a silent behavior change would go unnoticed. And
`From<PaymentAmountWire>` carried no note that the conversion is intentionally
unvalidated.

## Change (tests + one comment)

Added to the existing test module in `paykit-lib/src/payment_amount.rs`
(matching its existing unprefixed test naming):

- Asset containing a control character → rejected.
- `"1.2.3"` → rejected.
- `"."` → rejected.
- `test_payment_amount_bare_decimal_point_currently_accepted` — **pins** that
  `".5"` and `"10."` pass validation today, with a comment that tightening is
  a protocol decision; this test makes any such change a visible diff.

Added an audit comment on `From<PaymentAmountWire>`
(`paykit-lib/src/shared_wire.rs`): the conversion is intentionally
unvalidated; both current callers revalidate immediately after converting
(verified: `payment_request/wire.rs` via `PaymentRequestTerms::validate()`,
`receipt/wire.rs` via `validate_with_label`).

## Scope / impact

- **Tests and a comment only**; no production code. **Protocol impact: none.**

## Verification (local)

- `cargo clippy -p paykit-lib --all-targets --all-features -- -D warnings`
- `cargo test -p paykit-lib --lib payment_amount` — **7 passed**

### DIFF
--- a/paykit-lib/src/shared_wire.rs
+++ b/paykit-lib/src/shared_wire.rs
@@ -34,6 +34,12 @@ pub(crate) struct PaymentAmountWire {
     pub(crate) asset: String,
 }
 
+// audit: this conversion is intentionally unvalidated so wire deserialization
+// can carry raw strings across the parse boundary. Every caller revalidates
+// the resulting PaymentAmount: payment_request/wire.rs runs
+// PaymentRequestTerms::validate() right after the conversion, and
+// receipt/wire.rs calls validate_with_label("Receipt amount") immediately
+// after mapping. Keep it that way if new callers are added.
 impl From<PaymentAmountWire> for PaymentAmount {
     fn from(wire: PaymentAmountWire) -> Self {
         Self {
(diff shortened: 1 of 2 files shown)


paykit-lib/src/shared_wire.rs
```diff
@@ -34,6 +34,12 @@ pub(crate) struct PaymentAmountWire {
     pub(crate) asset: String,
 }
 
+// audit: this conversion is intentionally unvalidated so wire deserialization
```
ben-kaufman: I think this comment should be either removed or re-written to match the comments style of the codebase

Jasonvdb: Rewritten in a9ff3e5. Dropped the audit prefix, the caller list, and the imperative sign-off. Kept the one constraint the type system cannot express: the conversion is unvalidated and callers must validate the resulting `PaymentAmount`.

ben-kaufman (approved):

PR #112, opened 2026-07-16
fix(payment-request): reject inverted recurrence windows

Jasonvdb:
## Problem

Review asked for two changes: actually reject a `Recurrence` whose `ends_at` is
before `starts_at` instead of pinning the acceptance, and drop the
`BillingPeriod` inversion test as duplicate coverage.

## Change

- `Recurrence::validate` now rejects a non-null `ends_at` at or before
  `starts_at` ("Recurrence ends_at must be after starts_at"), mirroring the
  `BillingPeriod` check: parsed-timestamp comparison, `PaykitError::Validation`
  at the validate layer, `PaykitError::InvalidData` on the wire parse path.
  `anchor` remains unconstrained.
- `specs/payment-requests.md` gains the matching rule in the Recurrence
  section: "When `ends_at` is non-null, it MUST be after `starts_at`." The
  spec previously set no ordering rule, so the guard is a spec-level
  tightening; landing code and spec together keeps them in sync.
- The acceptance-pinning test is replaced by rejection tests (before and equal
  boundaries), a positive case, and a wire-path test asserting `InvalidData`
  through `parse_payment_request_json`.
- `test_billing_period_inverted_rejected` removed: duplicate of
  `payment_proof_rejects_invalid_billing_period_shape` (types.rs) and
  `payment_proof_rejects_invalid_billing_period_order` (wire.rs). The earlier
  claim in this PR that the BillingPeriod side was untested was wrong.
- The recurrence proptest generator now only emits valid windows (`ends_at`
  strictly after `starts_at`), since the serialize path validates terms.
- Doc note on the SDK/FFI mirror types (`PaymentRequestRecurrenceRecord`,
  `FfiPaymentRequestRecurrence`); no exported struct shapes change.

## Protocol impact

- Tightening: wire messages with inverted recurrence windows, previously
  accepted, now fail parse with `InvalidData`; constructing or serializing such
  a request fails with `Validation`. Validation lives in paykit-lib, so
  Swift/Kotlin bindings see the tightening through existing error surfaces.
- Replay and durability: live intake persists a now-invalid message as
  malformed-recognized without aborting the batch; derivation replay marks only
  the affected record invalid and continues. One edge for reviewer awareness:
  backup restore validates stored parse-status metadata against the current
  classifier, so a backup containing a stream item stored as valid with an
  inverted window would fail restore with a stale-parse-status error. This only
  affects backups that actually contain such an item.

## Verification (local)

- `cargo clippy -p paykit-lib --all-targets --all-features -- -D warnings` and
  `cargo clippy -p paykit-sdk --all-targets -- -D warnings`
- `cargo test -p paykit-lib --all-targets` (209 passed, includes the proptests)
  and `cargo test -p paykit-sdk --lib` (367 passed)
- `cargo doc --no-deps` clean

### DIFF
--- a/paykit-lib/src/payment_request/types.rs
+++ b/paykit-lib/src/payment_request/types.rs
@@ -96,7 +96,8 @@ pub struct Recurrence {
     pub starts_at: String,
     /// RFC3339 UTC timestamp using `Z`.
     pub anchor: String,
-    /// Optional RFC3339 UTC timestamp using `Z`.
+    /// Optional RFC3339 UTC timestamp using `Z`, after `starts_at` when
+    /// present.
     pub ends_at: Option<String>,
 }
 
@@ -527,16 +528,23 @@ impl PaymentRequestEventMessage {
 }
 
 impl Recurrence {
+    /// Check `every`, timestamp formats, and that a non-null `ends_at` is
+    /// after `starts_at`. `anchor` ordering is not constrained.
     pub(super) fn validate(&self) -> Result<()> {
         if self.every == 0 {
             return Err(PaykitError::Validation(
                 "Recurrence every must be a positive integer".into(),
             ));
         }
-        parse_utc_timestamp(&self.starts_at, "Recurrence starts_at")?;
+        let starts_at = parse_utc_timestamp(&self.starts_at, "Recurrence starts_at")?;
         parse_utc_timestamp(&self.anchor, "Recurrence anchor")?;
         if let Some(ends_at) = &self.ends_at {
-            parse_utc_timestamp(ends_at, "Recurrence ends_at")?;
+            let ends_at = parse_utc_timestamp(ends_at, "Recurrence ends_at")?;
+            if ends_at <= starts_at {
+                return Err(PaykitError::Validation(
+                    "Recurrence ends_at must be after starts_at".into(),
+                ));
+            }
         }
         Ok(())
     }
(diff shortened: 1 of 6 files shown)


paykit-lib/src/payment_request/types.rs
```diff
@@ -621,6 +621,63 @@ mod tests {
         assert!(matches!(err, PaykitError::Validation(ref msg) if msg.contains("Z suffix")));
     }
 
+    #[test]
+    fn test_recurrence_inverted_window_currently_accepted() {
```
ben-kaufman: We should reject a Recurrence when `ends_at` is before `starts_at`. This test instead makes the invalid range required to pass and locks in the current gap. Could we add the ordering check and change this to a rejection test, or at least remove the test that requires acceptance?

paykit-lib/src/payment_request/types.rs
```diff
@@ -621,6 +621,63 @@ mod tests {
...
+        let request = super::super::wire::parse_payment_request_json(json)
+            .expect("inverted Recurrence window is currently accepted");
+        let recurrence = request.request.recurrence.expect("Recurrence is present");
+        assert!(recurrence.validate().is_ok());
+        assert_eq!(recurrence.starts_at, "2026-07-01T00:00:00Z");
+        assert_eq!(recurrence.ends_at.as_deref(), Some("2026-06-01T00:00:00Z"));
+    }
+
+    #[test]
+    fn test_billing_period_inverted_rejected() {
```
ben-kaufman: This inversion is already covered by `payment_proof_rejects_invalid_billing_period_shape` in this file and `payment_proof_rejects_invalid_billing_period_order` in `wire.rs`. Could we remove this duplicate test and update the PR description's statement that this side was untested?

ben-kaufman (commented):
I left two inline comments.

paykit-lib/src/payment_request/types.rs
Jasonvdb: Added the ordering check in 8e24bf8. `Recurrence::validate` now rejects a non-null `ends_at` at or before `starts_at`, mirroring the `BillingPeriod` check: parsed-timestamp comparison, `Validation` from validate, `InvalidData` on the wire parse path. The pinning test is replaced with rejection tests (before and equal), a positive case, and a wire-path test in `wire.rs`. Since the spec set no ordering rule for Recurrence, the same commit adds the matching rule to `specs/payment-requests.md` so code and spec land together.

One thing to weigh at merge: backup restore re-classifies stored stream items and fails on stale parse-status metadata, so a backup containing an inverted-window request stored as valid would fail restore after this change. Live intake and derivation replay degrade per record without aborting. Details in the updated PR description.

paykit-lib/src/payment_request/types.rs
Jasonvdb: Confirmed and removed in 8e24bf8. Same validator, same fixture, same assertion: `payment_proof_rejects_invalid_billing_period_shape` covers it through `validate_for_request` and `payment_proof_rejects_invalid_billing_period_order` covers the wire path with the `InvalidData` remap. PR description updated to drop the claim that this side was untested.

paykit-lib/src/payment_request/wire.rs
```diff
@@ -763,6 +763,35 @@ mod tests {
         );
     }
 
+    #[test]
+    fn payment_request_rejects_invalid_recurrence_window_order() {
```
ben-kaufman, line 767: The new test covers inbound parsing, but not outbound serialization. Since rejection there depends on `serialize_payment_request_event` calling `validate`, could we add a regression test asserting `PaykitError::Validation` when `ends_at` is before or equal to `starts_at`?

ben-kaufman (commented):
I left one inline comment.

paykit-lib/src/payment_request/wire.rs
Jasonvdb: Added in 9e44879. `payment_request_rejects_outgoing_recurrence_ends_at_before_starts_at` and `payment_request_rejects_outgoing_recurrence_ends_at_equal_to_starts_at` build a full request and assert `PaykitError::Validation` from the serializer. They call `serialize_payment_request_json` directly: the `Request` arm of `serialize_payment_request_event` delegates straight to it and the `event.request.validate()` call lives there, matching the existing outbound test `payment_request_rejects_mutated_outgoing_kind`.

ben-kaufman (approved):

PR #113, opened 2026-07-17
Use receiver Noise keys for encrypted links

ben-kaufman:
## Summary

- publish a receiver-scoped Noise public key in each Receiver Marker and discover the counterparty key before starting an Encrypted Link
- require a separately persisted receiver Noise secret key for session construction, signup, signin, auth completion, and import across Rust, Swift, and Kotlin
- keep the Pubky identity secret optional so Ring- or server-owned identities can establish private links without exposing that secret to Paykit
- derive private and recovery paths with receiver Noise keys while retaining Pubky identity keys for homeserver routing and receiver-pair domain separation
- persist the counterparty receiver Noise public key in Encrypted Link snapshots so restored links do not repeat public discovery
- represent identity state with explicitly named local Pubky and receiver Noise public keys, plus matching live-session availability, without redundant capability enums or flags
- preserve cached private state and read-only access while live session access is temporarily unavailable
- persist the local receiver Noise public key as part of receiver identity, invalidate only private/link-derived state if it changes, and require backup restores to match both identity keys
- prevent reauthentication and session import from silently rotating receiver Noise keys
- zeroize receiver secret buffers and temporary conversion copies at the FFI boundary
- regenerate the Swift, Kotlin, iOS, and Android bindings and publish them as `0.1.0-rc37`

## Protocol impact

This uses the Receiver Marker introduced in #94. Paykit owns receiver-key discovery and path derivation, then injects the derived paths into `pubky-noise`; no `pubky-noise` API or wire change is required.

The unpublished Receiver Marker and Encrypted Link snapshot versions, path-domain labels, and Noise protocol version remain unchanged. No migration is required.

Identity status has three directly represented states:

- no persisted public key: explicit sign-out
- persisted public key without matching live session access: identity remembered, Pubky-backed workflows unavailable
- persisted public key with matching live session access: authenticated public and private workflows available

## Verification

- `cargo check --workspace --all-targets --all-features`
- `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --all-features --no-deps`
- `cargo test --workspace --all-features` through GitHub CI
- SDK unit tests: 370 passed
- FFI tests: 47 passed, 1 ignored; secret-wrapper integration tests: 4 passed
- SDK testnet E2E tests: 11 passed, including the server-owned identity case with no local Pubky identity secret
- `cd paykit-ffi && ./build.sh all` for iOS and every Android target, including Gradle/Maven-local validation
- `Package.swift` checksum matches the generated `Paykit.xcframework.zip`


paykit-sdk/src/runtime/mod.rs
dzdidi: Once an Encrypted Link exists, its local receiver Noise key must remain stable. A different Noise key effectively creates a different private receiver, even when the Pubky identity and receiver path are unchanged.
So this comparison is not enough, need a newly added key too (needs to be added to state).

dzdidi (approved):
LGTM

PR #114, opened 2026-07-20
Separate public and private payment flows

ben-kaufman:
## Summary

- split public and private receiving details, endpoint candidates, adapter callbacks, resolution statuses, and result types across the SDK and FFI
- remove the mixed contact-payment resolver, source discriminator, inclusion flag, and implicit private-to-public fallback
- expose independent public resolution, private resolution, and private preparation workflows
- add atomic Private Payment List freshness: private results return an opaque list version, and callers can require a version newer than the last consumed list
- return `WaitingForUpdatedPaymentList` without payable endpoints when the available private list is not newer
- regenerate Swift and Kotlin bindings plus all Android native libraries
- bump the workspace and platform packages to `0.1.0-rc39`

## Why

Public Payment Endpoints and counterparty-specific Private Payment Lists have different storage, privacy, linking, and recovery semantics. The SDK previously allowed them to be combined into one candidate batch and result, which let applications accidentally blur those boundaries.

This change makes the payment mode an explicit API choice at every layer. It also lets applications atomically prove which Private Payment List produced a resolution and avoid reusing any endpoint from a consumed list. A newer list is fresh as a whole, even when it intentionally repeats reusable Payment Endpoints.

## Developer impact

This is a breaking adapter and resolution API change. Integrators must implement the new public/private callback names and consume the corresponding source-specific result types. There is no compatibility shim or implicit fallback between payment modes.

Private resolver calls now accept optional `afterPrivatePaymentListVersion`, and private results expose `privatePaymentListVersion`. Applications should persist the returned version before submitting a payment, serialize that handoff per counterparty receiver, and treat submitted, pending, or uncertain payment execution as consuming the complete Private Payment List.

There is no Paykit Library wire-format change; freshness is SDK-local and exposed through the Swift and Kotlin bindings.

The `v0.1.0-rc39` prerelease is published from this branch so Bitkit integration can proceed during review.

## Validation

- `cargo fmt --all`
- `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --all-features --no-deps`
- `cargo test --workspace --all-features`
- `cd paykit-ffi && ./build.sh all` (Swift, four Android architectures, Kotlin generation, Maven-local publish check)


paykit-ffi/bindings/android/gradle.properties
```diff
@@ -3,4 +3,4 @@ android.useAndroidX=true
 android.enableJetifier=true
 kotlin.code.style=official
 group=com.synonym
-version=0.1.0-rc37
+version=0.1.0-rc39
```
dzdidi, line 6: jumed over

paykit-ffi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "paykit-ffi"
-version = "0.1.0-rc37"
+version = "0.1.0-rc39"
```
dzdidi, line 3: jumped over

paykit-lib/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "paykit-lib"
-version = "0.1.0-rc37"
+version = "0.1.0-rc39"
```
dzdidi, line 3: jumped over

paykit-sdk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "paykit-sdk"
-version = "0.1.0-rc37"
+version = "0.1.0-rc39"
```
dzdidi, line 3: jumped over

paykit-sdk/Cargo.toml
```diff
@@ -22,7 +22,7 @@ blake3 = "1"
 chrono = { version = "0.4.44", default-features = false, features = ["clock", "serde", "std"] }
 crypto_secretbox = "0.1"
 hex = "0.4"
-paykit-lib = { path = "../paykit-lib", version = "0.1.0-rc37" }
+paykit-lib = { path = "../paykit-lib", version = "0.1.0-rc39" }
```
dzdidi, line 25: jumped over

dzdidi (changes requested):

paykit-ffi/bindings/android/gradle.properties
ben-kaufman: Intentional, there was some issue in the Android release of 38 in the CI

dzdidi (approved):

PR #117, opened 2026-07-21
Document state blob store callback contract

Jasonvdb:
## Problem

Platform implementations of the SDK state blob store are synchronous foreign callbacks, but the public contract did not explain their blocking, concurrency, compare-and-swap, or sensitive-state requirements.

## Change

Document that:

- callbacks must perform prompt, bounded local persistence and must not depend on reentrant work from the same SDK handle;
- each SDK handle owns its transaction mutex, while revision reads and other handles can invoke callbacks concurrently;
- implementations must be thread-safe and any store-level coordination must remain bounded;
- loaded blobs and revisions must form a coherent snapshot;
- atomic saves use expected-revision semantics and leave stored data intact on failure;
- state blobs are not encrypted by Paykit and require protected storage, backup controls, and encryption before cloud or cross-device transport.

## Protocol impact

None. Documentation only.

## Public API/bindings impact

No UniFFI shape change. The documentation is carried into generated Swift and Kotlin bindings on the next release generation.

## Verification

- `RUSTDOCFLAGS="-D warnings" cargo doc -p paykit-ffi --no-deps`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test`
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps`
- `cd paykit-ffi && ./build.sh all` for Swift and all four Android ABIs
- `git diff --check`

A fresh isolated Cargo target directory was used instead of `cargo clean` to avoid shared-worktree stale artifacts. Generated release bindings and native libraries were used only to verify propagation and remain outside this documentation-only change.

### DIFF
--- a/paykit-ffi/src/storage.rs
+++ b/paykit-ffi/src/storage.rs
@@ -27,8 +27,10 @@ pub struct FfiSdkStateBlobSnapshot {
 ///
 /// Each store instance is bound to one complete logical SDK state record.
 /// Callbacks receive no identity or Paykit Receiver Path key, so apps must use
-/// separately namespaced store instances for different identities or configured
-/// Paykit Receiver Paths. Share a store between handles only when they
+/// separately namespaced store instances when identities or configured Paykit
+/// Receiver Paths need to retain distinct state. A store may be reused after
+/// clearing the previous identity's data when intentionally switching the
+/// logical state it represents. Share a store between handles only when they
 /// intentionally operate on the same logical state.
 ///
 /// Callbacks execute synchronously on the thread executing or polling the SDK
--- a/paykit-ffi/src/storage.rs
+++ b/paykit-ffi/src/storage.rs
@@ -24,15 +24,45 @@ pub struct FfiSdkStateBlobSnapshot {
 }
 
 /// Platform-owned durable blob store for SDK state.
+///
+/// Callbacks execute synchronously on the thread executing or polling the SDK
+/// call. Each SDK handle owns a transaction mutex shared by that handle's
+/// storage clones. The mutex serializes callbacks made inside SDK storage
+/// transactions on that handle; it does not coordinate the revision-read API,
+/// other SDK handles, or other processes.
+///
+/// Callbacks must return promptly. Use bounded local durable persistence and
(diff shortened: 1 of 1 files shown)


paykit-ffi/src/storage.rs
```diff
@@ -24,15 +24,45 @@ pub struct FfiSdkStateBlobSnapshot {
...
+///
+/// Callbacks must return promptly. Use bounded local durable persistence and
+/// bounded local coordination needed for atomic compare-and-swap. Do not use
+/// network or iCloud round trips, unbounded cross-process waits, or re-enter or
+/// depend on work from the same SDK handle.
+///
+/// A slow transactional callback blocks that foreign thread and queues other
+/// storage transactions on the same SDK handle. The revision-read API invokes
+/// the load callback outside the transaction mutex and can overlap transaction
+/// callbacks. Other SDK handles can also invoke callbacks concurrently.
```
ben-kaufman, line 50: Could we clarify the storage scope here? These callbacks do not receive a handle or state key, so one store instance represents one complete logical SDK state record. Handles for different identities or Paykit Receiver Paths need separately namespaced store instances. Multiple handles should share a store only when they intentionally share that same state.

ben-kaufman (commented):
One storage scope clarification.

paykit-ffi/src/storage.rs
Jasonvdb: Fixed in `07daa2f`. The store docs now clarify that each store instance represents one complete logical SDK state record. Different identities or Paykit Receiver Paths need separately namespaced stores, and handles should share a store only when they intentionally share the same state.

ben-kaufman: Thanks, this is clearer. One small qualification: the SDK can switch the active identity in the same state record by clearing the old identity's data and reusing the store. Could we say separate stores are required only when the identities or Paykit Receiver Paths need to keep distinct state? That still covers separate accounts without ruling out the supported identity switch.

Jasonvdb: Updated in `16ede13`. The docs now require separate stores only when identities or Paykit Receiver Paths need to retain distinct state, and clarify that a cleared store can be reused for an intentional identity switch.

PR #130, opened 2026-08-04
feat: expose local Pubky testnet config

ben-kaufman:
## Summary

- keep the public Pubky network as the default binding configuration
- use one optional `localTestnetHost` / `local_testnet_host` field to opt into local testnet
- regenerate the Swift, Kotlin, and Android native bindings for every supported target

## Protocol impact

None. This changes client transport configuration only; wire formats, storage paths, message kinds, and capability strings are unchanged.

## Downstream bindings

`FfiPubkyClientConfig` retains the existing timeout and adds one optional local-testnet host. `None` uses production; a configured host uses the standard local-testnet ports. No environment enum or additional configuration type is required.

## Release

- published `v0.1.0-rc42`: https://github.com/pubky/paykit-rs/releases/tag/v0.1.0-rc42
- Swift XCFramework checksum: `1223552e563ab1c6fc1fd0dc3e830e412c7a29cabf9b7bc91384ef9651ea4766`
- Android package publication is verified by the release workflow

## Checklist

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo test --workspace --all-features`
- [x] `cargo doc --workspace --no-deps --all-features`
- [x] Platform bindings rebuilt together with `cd paykit-ffi && ./build.sh -r --rc all`
- [x] Swift generated bindings type-check
- [x] Android release AAR builds and publishes to Maven Local


paykit-ffi/src/session.rs
```diff
@@ -578,6 +578,9 @@ pub(crate) fn pubky_from_config(config: &FfiPubkyClientConfig) -> Result<Pubky,
     }
 
     let mut builder = PubkyHttpClient::builder();
+    if let Some(host) = config.local_testnet_host.as_deref() {
+        builder.testnet_with_host(host);
```
ovitrif, line 583: Any `Some(local_testnet_host)` is passed straight into `PubkyHttpClientBuilder::testnet_with_host`, and Pubky panics on hosts that yield an invalid relay URL (via `.expect("relays urls infallible")`), so values like `""`, `"not a host"`, or unbracketed IPv6 can abort the process at the FFI boundary instead of returning a validation error. a59d7e0 previously rejected those inputs before calling Pubky. Could we restore that Domain/IPv4 validation (and the invalid-host unit test) before calling `testnet_with_host`?

ovitrif (commented):
Unvalidated `local_testnet_host` values can crash the native Paykit library during client construction instead of returning a recoverable validation error.

paykit-ffi/src/session.rs
ben-kaufman: Fixed in a9397f4. The FFI config conversion now validates the override as a nonempty Domain or IPv4 host before calling testnet_with_host, and returns a recoverable validation error for empty, whitespace-prefixed, malformed, or unsupported hosts. The invalid-host unit coverage is restored. The fix is included in rc43, and the PR format, clippy, tests, and docs checks pass.

ovitrif (approved):
Confirmed the local testnet host validation fix; invalid hosts now return a recoverable error instead of panicking at the FFI boundary.

Jasonvdb (approved):

